PR #5, opened 2022-04-08
Dataset config

valsdav:
Reorganized dataset handling.
Now the analysis cfg file contains just the list of the datasets json you want to run on and a filter by sample/year.
    
The dataset definition used by the build_dataset script are saved in a new json file.


mmarchegiani (approved):
Very handy feature for managing and filtering datasets "on the fly", without touching the json datasets

PR #19, opened 2022-06-09
Implementation of JER smearing with hybrid method

mmarchegiani:
The **JER smearing** with the hybrid method is implemented, following the recommendations on the corresponding [CMS Twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution). The method is a combination of a deterministic scaling method and a stochastic smearing.
The AK4 jets are **matched to a gen-jet**, if the requirements are met. The 4-momentum of matched jets is corrected by a factor dependent on the transverse momentum difference $p_T - p_{T,gen}$ and the data-to-simulation core resolution scale factor $s_{JER}$.
The 4-momentum of **unmatched jets** is instead corrected by a stochastic factor dependent on the scale factor $s_{JER}$ and the energy resolution $\sigma_{JER}$.
The energy resolution and scale factors are read from the JSON files recommended by the newest [JSON-POG scheme](https://cms-nanoaod-integration.web.cern.ch/commonJSONSFs/JME_jet_jerc_Run2_UL/).
In order to enforce **reproducibility** of the analysis, the random number generator is given a seed as input that is unique for each processed chunk. In this way, if the analysis is rerun with the same configuration and chunk size, the same stochastic smearing is applied and therefore the same results are reproduced. The per-chunk seed is also saved in the output file for bookkeeping.

### DIFF
--- a/notebooks/jer_test.ipynb
+++ b/notebooks/jer_test.ipynb
@@ -433,8 +433,141 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 15,
    "id": "da30c6f9",
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "<Array [10, 9, 10, 8, 5, 7, ... 6, 8, 9, 8, 9] type='10000 * int64'>"
+      ]
+     },
+     "execution_count": 15,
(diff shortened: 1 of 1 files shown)


valsdav:
Good for me, thanks!

PR #20, opened 2022-06-14
Implementation of semileptonic selection and triggers

mmarchegiani:
The semileptonic selection is implemented. A dedicated library is defined to apply the triggers, where the trigger application is parametrized by the final state.
The semileptonic triggers are updated accordingly for the 2018UL dataset. A future PR will include also the triggers for the 2016UL and 2017UL datasets.
The class `Dataset` is updated to handle the output of the `dasgoclient` command and save the file list correctly.

Additionally, a new JSON scheme is implemented to build the dataset and bookkeep the samples' names and eras.

Finally, `Configurator` is modified such that the script `scripts/plot/make_plots.py` will plot the histograms reading the most recent version of the output, unless specified otherwise.


PocketCoffea/parameters/lumi.py
```diff
@@ -1,3 +1,10 @@
+# Function to convert and round the integrated luminosity from picobarn to femtobarn
```
valsdav, line 1: Is this really necessary? Ok

PocketCoffea/parameters/triggers.py
```diff
@@ -65,5 +65,38 @@
...
+                "Ele28_eta2p1_WPTight_Gsf_HT150"
+            ],
+            "mu" : [
+                "IsoMu24_eta2p1_TightChargedIsoPFTauHPS35_Trk1_eta2p1_Reg_CrossL1",
+                "IsoMu24_eta2p1_MediumChargedIsoPFTauHPS35_Trk1_TightID_eta2p1_Reg_CrossL1",
+                "IsoMu24_eta2p1_TightChargedIsoPFTauHPS35_Trk1_TightID_eta2p1_Reg_CrossL1",
+                "IsoMu24_eta2p1_MediumChargedIsoPFTauHPS35_Trk1_eta2p1_Reg_CrossL1",
+                "IsoMu24",
+                "IsoMu24_eta2p1",
+                "IsoMu24_TwoProngs35",
```
valsdav, line 97: I'm not sure including everything is the best choice. For example the TwoProngs triggers I don't know what it does.

scripts/plot/make_plots.py
```diff
@@ -77,16 +79,26 @@
 
 selection = {
     'trigger'  : (r'Trigger'),
-    'baseline' : (r'Trigger'+'\n'+
-                  r'Dilepton cuts')
+    'dilepton_SR' : (r'Trigger'+'\n'+
```
valsdav, line 82: This is hardcoded in the plotting code.. we will need to remove it in a next PR.

PocketCoffea/parameters/lumi.py
mmarchegiani: Maybe it's over-engineered, but it's just to keep track of the reason why we divide the luminosity by a factor 1000.

PocketCoffea/parameters/triggers.py
mmarchegiani: Sure, we will have to check which triggers need to be applied.

PR #23, opened 2022-06-29
Semileptonic electron trigger efficiencies and SF

mmarchegiani:
With this PR, the single electron trigger SF computation is implemented. A dedicated processor `semileptonicTriggerProcessor` derived from the base processor is created for the purpose, that on top of the basic tasks is saving the 2D histograms to evaluate the trigger efficiencies and SF as a function of ($p_T$, $\eta$).
In order to be sure that a restricted set of weights is applied to events for this measurement, the method `compute_weights()` is overwritten in the trigger processor. Additionally, dedicated preselections on objects, trigger requirements and event selection are developed to select events.

To produce the 2D efficiency and SF maps, a dedicated script `scripts/plot/trigger_efficiency.py` is created, that saves directly in the output directory the relevant histograms.

In a future PR, also 1D histograms will be produced to show the efficiency as a function of $p_T$, $\eta$ and  $\phi$.

The PR also solves a bug in the normalization of MC. While before the MC yields were mistakenly divided by the sum of the weights progressively during processing, now the histograms are normalized only once in the `post_process()`, leading to the correct result for the MC normalization.


valsdav:
Changes are understood. 

Many improvements. This should also fix the problem with categories without events in that chunk.

PR #24, opened 2022-07-05
Parton matching now working with Numba compiled functions on dask

valsdav:
Fixed problems with numba functions: now they get compiled only once. 

Running on Dask works well now.


mmarchegiani (approved):
Hopefully this provides a final solution to the memory leak in the parton matching processor. Well done!

PR #30, opened 2022-07-13
Trigger SF plot with systematic variations

mmarchegiani:
A dedicated library and script are implemented to compute the data and MC trigger efficiencies and SF, and produce plots that are stored in the output directory.
The SF are plotted with their statistical uncertainty and the systematic variations for given systematic effects are also evaluated and plotted together with the nominal value.
In a future PR also the difference between the Up/Down variations and the nominal SF value will be included in the plots.


PocketCoffea/lib/fill.py
```diff
@@ -29,16 +29,19 @@ def fill_histograms_object_with_variations(processor, obj, obj_hists, systematic
...
+
+        for (var, modifier) in modifiers.items():
+            for category, cuts in processor._categories.items():
+                if event_var:
+                    weight = processor.weights.weight(modifier=modifier) * processor._cuts_masks.all(*cuts)
+                    fields = {k: ak.fill_none(getattr(processor.events, k), -9999) for k in h.fields if k in histname}
+                else:
+                    weight = ak.flatten( processor.weights.weight(modifier=modifier) * ak.Array(ak.ones_like(obj.pt) *
+                                                            processor._cuts_masks.all(*cuts)) )
+                    fields = {k: ak.flatten(ak.fill_none(obj[k], -9999)) for k in h.fields if k in dir(obj)}
```
valsdav, line 46: the improvement for the fill_none case should be included here

PocketCoffea/utils/Plot.py
```diff
@@ -0,0 +1,332 @@
+import os
```
valsdav, line 1: This seems related only to the SF plots. Why it is part of the utils/Plot.py file?

scripts/plot/trigger_efficiency.py
```diff
@@ -16,9 +16,9 @@
 
 from multiprocessing import Pool
 from PocketCoffea.parameters.allhistograms import histogram_settings
-from PocketCoffea.parameters.lumi import lumi, femtobarn
 
 from PocketCoffea.utils.Configurator import Configurator
+from PocketCoffea.utils.Plot import plot_efficiency
```
valsdav, line 21: Ah ok! It is a general efficiency plotting macro!

PocketCoffea/utils/Plot.py
mmarchegiani: Several functions are defined in `utils/Plot.py`, not only related to the SF plots but also for the efficiency + uncertainty computation and plot of the efficiency (1D + 2D plots). These functions could be used not only for the trigger efficiency, but for any kind of efficiency that we want to compute and/or plot, therefore the idea to define a module in `utils`.
However the location of the module is not binding, we could move it in another folder if preferable.

PR #31, opened 2022-08-08
Semileptonic trigger SF computation

mmarchegiani:
The semileptonic trigger SF are computed in a (pt, eta) grid and saved as a `dense_lookup` 2D table.
The trigger SF are applied in the base processor by activating a flag.
A preliminary implementation of the eras splitting is also included and will be ultimated in a next PR.


PocketCoffea/lib/fill.py
```diff
@@ -42,14 +45,26 @@ def fill_histograms_object_with_variations(processor, obj, obj_hists, systematic
                 if event_var:
                     keys = [k for k in h.fields if k in histname]
                     isnotnone = ~ak.is_none(getattr(processor.events, keys[0]))
-                    weight = processor.weights.weight(modifier=modifier) * processor._cuts_masks.all(*cuts)[isnotnone]
+                    if triggerSF:
```
valsdav, line 48: We will need to find a way to generalize this. Like passing an array of custom weights.

PR #39, opened 2022-11-09
Migration of semileptonic trigger SF library to hist

mmarchegiani:
The present PR re-implements the semileptonic trigger SF library in order to migrate from `Coffea` histograms to `hist` histograms.

### DIFF
--- a/PocketCoffea/utils/PlotEfficiency.py
+++ b/PocketCoffea/utils/PlotEfficiency.py
@@ -397,24 +397,27 @@ round_opts = {
 }
 
 def dense_dim(h):
+    '''Returns the number of dense axes of a histogram.'''
     dense_dim = 0
     if type(h) == dict:
         h = h[list(h.keys())[0]]
     for ax in h.axes:
-        if not type(ax) == hist.axis.StrCategory:
+        if not type(ax) in [hist.axis.StrCategory, hist.axis.IntCategory]:
             dense_dim += 1
     return dense_dim
 
 def dense_axes(h):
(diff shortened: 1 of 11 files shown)


PocketCoffea/utils/PlotEfficiency.py
```diff
@@ -6,8 +6,9 @@
 
 import mplhep as hep
 import hist
+from hist import Hist
 from coffea import processor
-from coffea.hist import Bin, Hist, plot
+#from coffea.hist import Bin, Hist, plot
```
valsdav, line 11: Let's delete completely the outdated lib

PocketCoffea/workflows/base.py
```diff
@@ -116,7 +109,7 @@ def load_metadata(self):
         self._year = self.events.metadata["year"]
         self._triggers = triggers[self.cfg.finalstate][self._year]
         self._btag = btag[self._year]
-        self._isMC = self.events.metadata["isMC"]
+        self._isMC = self.events.metadata["isMC"] == "True"
```
valsdav, line 112: Is this a string at the end in the dataset definition?

valsdav: Probably better to be sure in build_dataset to have a boolean..

valsdav (commented):
The changes are fine for me!

PocketCoffea/workflows/base.py
mmarchegiani: Yes, at the moment the attribute `isMC` is a string. This needs to be fixed when building the json dataset.

PR #60, opened 2023-02-21
Trigger SF and b-tag calibration

mmarchegiani:
The present PR includes the workflow used to compute the single electron trigger Scale Factors as 2D maps in ($\eta_{SC}$, $p_T$) of the electron and the updated b-tag SF calibration. The code has been tested on a 2018 UL dataset and there will be a future update including the 2017 and 2016 UL datasets.
### Single electron trigger SF
The final SF maps are saved in `pocket_coffea/parameters/semileptonic_triggerSF`. The SF weights are now available to be included from the config file with the keyword `sf_ele_trigger`. Multiple variations are stored in the weight, corresponding to the statistical uncertainty on the SF and the systematic uncertainties corresponding to the pileup reweighting, the era dependence and the difference between low/high $H_T$ events.
Several tests have been performed to ensure closure after the SF application, by checking the electron observables 1D distributions before and after the application of the trigger SF, and a general improvement of data/MC agreement was observed.
Additionally, the 2D map in ($\phi$, $\eta_{SC}$) of the electron was investigated before/after the application of the Scale Factor to check if the detector response is well modeled.
### B-tag SF calibration
The b-tag SF calibration has been updated to account for additional background (single-top) and is computed **after** application of **all the lepton and trigger SF**.
A dedicated closure test has been performed to ensure that after the application of the b-tag SF and the b-tag calibration SF the distributions are not changed in shape with respect to the distributions before application of the b-tag SF.


config/semileptonic_triggerSF/plot_options.py
```diff
@@ -1,343 +1,343 @@
 efficiency = {
     '2016_PreVFP' : {
         'Ele32_EleHT_pass': {
-            'hist_electron_pt' : {'ylim' : (0, 1)},
```
valsdav, line 4: Why is this dictionary necessary? It would be best to read the histogram config from the configuration. This is only replication, no?

notebooks/parton_matching_bpartons.ipynb
```diff
@@ -1203,7 +1203,7 @@
    "name": "python",
```
valsdav, line 1203: Please remove this file from the commit.

pocket_coffea/utils/plot_utils.py
```diff
@@ -200,65 +301,149 @@ def get_systematic_uncertainty(
...
+                syst_err2_up_twosided = np.where(up_is_up, errUp**2, errDown**2)
+                syst_err2_down_twosided = np.where(up_is_up, errDown**2, errUp**2)
+                syst_err2_max = np.maximum(syst_err2_up_twosided, syst_err2_down_twosided)
+                syst_err2_up_onesided = np.where(is_onesided & up_is_up, syst_err2_max, 0)
+                syst_err2_down_onesided = np.where(is_onesided & down_is_down, syst_err2_max, 0)
+                syst_err2_up_combined = np.where(is_onesided, syst_err2_up_onesided, syst_err2_up_twosided)
+                syst_err2_down_combined = np.where(is_onesided, syst_err2_down_onesided, syst_err2_down_twosided)
+
+                syst_err2_up_dict[syst] += syst_err2_up_combined
+                syst_err2_down_dict[syst] += syst_err2_down_combined
```
valsdav, line 321: Ok very nice! This should be a much more consistent implementation of prefit uncertainty band plot.

pocket_coffea/utils/plot_utils.py
```diff
@@ -53,6 +137,17 @@
     'bb': (0.51, 0.91, 0.51),  # green
 }
 
+colors_tthbb = {
```
valsdav, line 140: All of this needs to be part of the configuration of the particular analysis run, not of the plotting script... to be done in a separate PR

pocket_coffea/utils/plot_utils.py
```diff
@@ -200,65 +301,149 @@ def get_systematic_uncertainty(
...
+
+    h_mc_sum = stack_sum(stack_mc_nominal)
+    nom = h_mc_sum.values()
+
+    if split_systematics and only_syst:
+        syst_err2_up_filtered = { syst : unc for syst, unc in syst_err2_up.items() if any(key in syst for key in only_syst)}
+        syst_err2_down_filtered = { syst : unc for syst, unc in syst_err2_down.items() if any(key in syst for key in only_syst)}
+
+    if not split_systematics:
+        plot_uncertainty_band(h_mc_sum, syst_err2_up, syst_err2_down, ax, ratio)
```
valsdav, line 377: this is not using the options `only_syst`. 
I think that the function must be coherent with the argument. If `only_syst` is given, also the plot with the total uncertainty should only contain the filtered systematic.

pocket_coffea/utils/plot_utils.py
```diff
@@ -395,6 +580,13 @@ def plot_data_mc_hist1D(
             else:
                 dict_mc = {d: h[d][slicing_mc] for d in samples_mc}
                 dict_mc_nominal = {d: h[d][slicing_mc_nominal] for d in samples_mc}
+                nevents = {
+                    d: round(sum(dict_mc_nominal[d].values()), 1) for d in samples_mc
```
valsdav, line 584: The number of events should not be rounded to ingeer MC events, but at least rounded to to have 1 decimal digit.

valsdav (changes requested):

pocket_coffea/utils/plot_utils.py
mmarchegiani: This is what `round(sum(dict_mc_nominal[d].values()), 1)` is actually doing. When the second argument of `round()` is set to 1, the number is rounded to the first decimal digit. If the argument was instead set to 0, that would correspond to an integer rounding, but it is not the case.

pocket_coffea/utils/plot_utils.py
mmarchegiani: Sure, this will be moved to the configuration file in a following PR featuring also a refactorization of the plotting code.

config/semileptonic_triggerSF/plot_options.py
mmarchegiani: The reason for this dictionary to exist is to be able to configure the y-axis limits in data and MC trigger efficiency plots for each observable separately. These parameters are not directly related to the histogram itself but on the trigger efficiency as a function of the variable. At the moment `ylim` is set to a dummy value of `(0, 1)` but in the future it can be tuned to improve each plot.

pocket_coffea/utils/plot_utils.py
mmarchegiani: This is fixed with a patch commit: https://github.com/PocketCoffea/PocketCoffea/pull/60/commits/2e3e318a91165114535b818139a84feea627eadd

pocket_coffea/utils/plot_utils.py
valsdav: Ok thanks, I missed the argument of `round()` :)

PR #61, opened 2023-03-06
BugFix for btagSF variation

valsdav:
The variation of the btagSF were filled in the wrong way because of the c-flavour masks.

The light flavour variations were filled with 1 instead of the **nominal btagSF** when the `cferr*` variations were computed. 

This was causing one-sided uncertainty bands for cferr variations. 

N.B.: Implemented the btagSF central value for JES variations.


mmarchegiani (approved):
Good bugfix that solves the problem of the `sf_btag_cferr*` systematic uncertainties appearing as one-sided uncertainties. The bug is limited to the b-tag SF variations and is not affecting the b-tag SF calibration and single electron trigger SF derivation.

PR #62, opened 2023-03-07
Sum of genweights as a dataset metadata parameter

mmarchegiani:
With the present PR we introduce a new scheme for rescaling MC samples by the sum of genweights. Instead of computing the sum of the genweights each time the `ttHbbBaseProcessor` is run and rescale MC in `postprocess()`, the sum of genweights are computed with a separate processor `genWeightsProcessor` once for all and included in the datasets definitions as an additional metadata parameter.

These changes allow to label different datasets to belong to the same sample by keeping track of their cross-section and sum of genweights through the metadata. Additionally, an extra key `part` is added to the metadata to specify which part of the sample the dataset belongs to (e.g. $H_T$ bin, s-/t-channel, ...).


pocket_coffea/lib/weights_manager.py
```diff
@@ -139,6 +139,7 @@ def __init__(
         self._year = metadata["year"]
         self._finalstate = metadata["finalstate"]
         self._xsec = metadata["xsec"]
+        self._sum_genweights = float(metadata["sum_genweights"])
```
mmarchegiani, line 142: For some reason, unlike `metadata["xsec"]` the parameter `metadata["sum_genweights"]` is a string so it needs to be casted into a float.

pocket_coffea/lib/weights_manager.py
```diff
@@ -229,7 +230,7 @@ def _compute_weight(self, weight_name, events, shape_variation):
         in the constructor.
         '''
         if weight_name == "genWeight":
-            return [('genWeight', events.genWeight)]
+            return [('genWeight', events.genWeight / self._sum_genweights)]
```
mmarchegiani, line 233: Each event is weighted by the `genWeight` divided by the sum of the `genWeights`.

pocket_coffea/workflows/base.py
```diff
@@ -767,6 +773,7 @@ def postprocess(self, accumulator):
                         continue
                     h *= scale_genweight[sample]
             accumulator["scale_genweight"] = scale_genweight
+            """
```
mmarchegiani, line 776: At the moment this part is just commented it out but could be removed.

pocket_coffea/utils/accumulate.py
```diff
@@ -0,0 +1,63 @@
...
+    
+    return accumulator
+
+def join_accumulator(accumulator, joint_dataset):
+    '''This function joins the total accumulator into the accumulator of a single dataset.'''
+    for key in ['sumw', 'cutflow']:
+        for k, d in accumulator[key].items():
+            accumulator[key][k] = {joint_dataset : sum(d.values())}
+
+    for key in ['sum_genweights']:
```
valsdav, line 23: With the new implementation of the `sum_genweights`,  this should be remove from the accumulation process.

pocket_coffea/workflows/base.py
valsdav: No dangling comment code..

pocket_coffea/workflows/genweights.py
```diff
@@ -0,0 +1,53 @@
...
+
+class genWeightsProcessor(BaseProcessorABC):
+    def __init__(self, cfg: Configurator) -> None:
+        super().__init__(cfg)
+        self.output_format = {
+            "sum_genweights": {},
+            "sumw": {
+                cat: {} for cat in self._categories
+            },
+            "cutflow": {
```
valsdav, line 16: Cutflow dictionary needs only the `initial` key

pocket_coffea/workflows/genweights.py
```diff
@@ -0,0 +1,53 @@
+import awkward as ak
+
+import copy
+
+from .base import BaseProcessorABC
+from ..utils.configurator import Configurator
+
+class genWeightsProcessor(BaseProcessorABC):
```
valsdav, line 8: No need to derive it from the BaseProcessorABC, we are not using any facility of it

pocket_coffea/workflows/genweights.py
```diff
@@ -0,0 +1,53 @@
...
+            "sumw": {
+                cat: {} for cat in self._categories
+            },
+            "cutflow": {
+                "initial": {},
+                "skim": {},
+                "presel": {},
+                **{cat: {} for cat in self._categories},
+            },
+            "variables": {
```
valsdav, line 22: Remove all this fields that are not used

pocket_coffea/workflows/genweights.py
valsdav: Or do you want to use some function of it? OK

scripts/append_genweights.py
```diff
@@ -0,0 +1,43 @@
...
+import json
+import numpy as np
+from coffea.util import load
+
+from pocket_coffea.utils.dataset import Dataset
+
+parser = argparse.ArgumentParser(description='Build dataset fileset in json format')
+parser.add_argument(
+    '--cfg',
+    default=os.getcwd() + "/datasets/datasets_definitions.json",
```
valsdav, line 12: Remove this default, better to be sure where we are applying it

pocket_coffea/workflows/genweights.py
mmarchegiani: The idea is to use the same dict structure and same metadata parameters as defined in the `BaseProcessorABC`.

valsdav (approved):

PR #67, opened 2023-03-23
Update genweights processor and hadd script

mmarchegiani:
The present PR updates the `genWeightsProcessor` to make it compatible with the skimming.
In the `cutflow` dictionary, the values `initial` and `skim` correspond to the number of events before/after the skim for each dataset (unlike in the `BaseProcessorABC` where the dictionary is by sample).

In addition, the script that performs the "hadd" of the skimmed output is updated and documented.

Minor modifications in the plotting code are included, but will be followed by a dedicated PR for the refactoring of the plotting code.


pocket_coffea/workflows/genweights.py
```diff
@@ -48,4 +94,13 @@ def process(self, events: ak.Array):
             if self._hasSubsamples:
                 raise Exception("This processor cannot compute the sum of genweights of subsamples.")
 
+        self.process_extra_before_skim()
+        self.skim_events()
```
valsdav, line 98: This call should be moved to line 104. No need to skim in the genweights processor if not for saving the skimmed files.

PR #68, opened 2023-03-24
Use MaskStorage for subsamples cuts

mmarchegiani:
The present PR implements the use of the `MaskStorage` object to store the subsamples masks.
Now the configuration to define the subsamples follows the same scheme as the categories. For each sample, one can specify the subsamples as a dictionary, a `StandardSelection` or a `MultiCut`. The `Configurator` module will take care of converting the dictionary into a `StandardSelection`. The dictionary of subsamples returned by the `Configurator` is effectively a dictionary of dictionaries whose values are `StandardSelection` and `MultiCut` objects.
The new implementation has been tested on a configuration with a categorization per object and a definition of subsamples per object. It was checked explicitly that the subsamples are properly defined when they are defined per object.


valsdav:
+1. This simplifies  the subsample configuration and the base workflow code. Thanks

PR #73, opened 2023-04-17
New parameters setup and Configurator improvements

valsdav:
- Created default parameters in yaml format to be composed with OmegaConf
- Changing Configurator to make it a standalone object separated from the running configuration


valsdav:
Merging this for further testing after tagging the previous version with 0.1

PR #75, opened 2023-04-20
Analysis example

valsdav:


mmarchegiani:
We merge the changes made following the CAT Hackathon with the updated documentation including a full analysis example.

PR #91, opened 2023-08-25
Include 2016UL datasets

mmarchegiani:
The datasets in `datasets/datasets_definitions.json` are **updated** to include the 2016UL datasets, including the pre/post-VFP eras. The file is renamed as `datasets_definitions_example.json`, since this file only provides an example of datasets definition, that the users can use as a template to write their custom datasets definition.

The .json datasets generated with `build_datasets.py` are **removed** from the `datasets/` folder as this is analysis specific metadata that is instead moved to the `AnalysisConfigs` repository.

The list of updated datasets is the following:

- SingleEle
- SingleMu
- TTTo2L2Nu
- TTToSemiLeptonic

In this new version of the datasets, the `sumgenweights` parameter is removed from the dataset metadata, since in the newest version of PocketCoffea the sum of genweights is computed "on the fly" by the processor and is not needed as an input parameter anymore.

Additionally, a short guide on the usage of the `build_dataset.py` script is included as a README file.


andreypz:
In my opinion, the `datasets` directory should not be a part of PocketCoffea repo. It should go to repo for Analysis Config.
The datasets local redirector depends on the user, so the files in github would be useless for other people.
Instead there could be a short README file that gives commands to re-produce those files.
On the other hand the  `datasets_definision.json` should probably stay, and contain latest version of UL/Run3/Other datasets (better as separate files).

mmarchegiani:
@andreypz as you suggest the datasets don't need to be in the PocketCoffea repository, as this is analysis specific metadata and has to be stored in the corresponding analysis folder in the `AnalysisConfigs` repo.
In a following PR we will clean this folder and keep only the UL datasets definitions.

PR #97, opened 2023-09-12
Modifications need for workflows to work at RWTH

andreypz:
For now this is just work in progress modifications I had to do to make it work for me. To keep track of them.
Once finalized, I will write a more detailed summary.


andreypz:
Please consider this PR for merging.
Here is the list of changes in order of appearance:
* `jets.py`: introducing dijet object (for events with >=2 good jets). In case of Njets>2 - take the first two.
* `leptons.py`: rewrote it by combining muon and electron collections into one. This simplifies the code.
* `triggers.py`: must check if the trigger bit exists before asking for it! In data some bits appear in the middle of data-taking and the bits do not exist in earlier runs.
* `histograms.py`: changed the ranges of some histograms, to what I think makes more sense for general analysis. Note: one needs to write a tutorial chapter on how to define your own histograms outside of the main pocket_coffea repository. But also with a possibility to modify the ones that are from pocket_coffea (like binning or labels)
* `datasets.py`: just added some urllib3 lines  to avoid  warnings from rucio.
* `plot_utils.py`: option to ignore error messages from numpy (coming from dividing histograms with empty bins probably)
* `rucio.py`: for the cases when local username is not the same as CERN username (running not on lxplus), one has to use the CERN username for Rucio to work. One has to set local environment variable `CERN_USERNAME`,  which is picked up by the script. If not set (default), then local username is used. Should be added to the manual as well.
* `base.py`: fix to avoid crashes due to missing parameter in config
* `runner.py`: modifications needed to Parsl submission to work on RWTH Linux cluster with Condor (should not affect dask submissions on other sites)

valsdav:
Hi @andreypz ! Thanks a lot for the PR, I'm going throught it and I think we can merge it very soon!

pocket_coffea/lib/leptons.py
```diff
@@ -42,32 +42,19 @@ def get_dilepton(electrons, muons, transverse=False):
...
-    electrons = ak.pad_none(electrons, 2)
-    muons = ak.pad_none(muons, 2)
-
-    nelectrons = ak.num(electrons[~ak.is_none(electrons, axis=1)])
-    nmuons = ak.num(muons[~ak.is_none(muons, axis=1)])
-
-    ee = electrons[:, 0] + electrons[:, 1]
-    mumu = muons[:, 0] + muons[:, 1]
-    emu = electrons[:, 0] + muons[:, 0]
+    leptons = ak.pad_none(ak.with_name(ak.concatenate([ muons, electrons], axis=1), "PtEtaPhiMCandidate"), 2)
```
valsdav, line 45: I agree that defining leptons here make the second part of the function easier. 
For performance sake, would it make sense to concatenate maximum 2 objects per collection as it was done before? 

Can you check with a quick profiling using `--test --n-chunks 10` if padding the full collection of just 2 elements for ele and for mu does change the performance?

pocket_coffea/lib/triggers.py
```diff
@@ -55,7 +55,8 @@ def get_trigger_mask(events, trigger_dict, year, isMC, primaryDatasets=None, inv
             )
             trigger_mask = trigger_mask | (events.HLT[trigger] & flag)
         else:
-            trigger_mask = trigger_mask | events.HLT[trigger.lstrip("HLT_")]
+            if trigger in events.HLT.fields:
```
valsdav, line 58: I think that here we are missing an expection: in case the trigger is not present in the fields,  this would fail  __silently__.  Instead without this if, coffea would emit an exception about a missing filter and the processing would stop. 

I think we can avoid the if and relying on the clear expection in the case of missing trigger.

scripts/runner.py
```diff
@@ -126,22 +126,27 @@ def load_run_options(cfg):
             'export XRD_RUNFORKHANDLER=1',
             f'export X509_USER_PROXY={_x509_path}',
             # f'export X509_CERT_DIR={os.environ["X509_CERT_DIR"]}',
-            'source /etc/profile.d/conda.sh',
+            'source /etc/profile.d/conda.sh',  # This looks site-specific, may not work everywhere.
```
valsdav, line 129: true, we need still to refactor the runner interface to make it more sites-generic.

valsdav (changes requested):
Thanks a lot for all the work!

pocket_coffea/lib/leptons.py
andreypz: I did some tests and the difference runtime is not even noticeable. Most of the events have two muons or less anyway. But I changed the code to consider only up to two muons/electrons.

scripts/runner.py
andreypz: The way to unify this, I think, is to execute `.bashrc`. Then it's the responsibility of the user to setup conda there.

pocket_coffea/lib/triggers.py
valsdav: Ah ok I see the problem. But triggers are defined by data taking period. 
Are those triggers present only on part of the NanoAOD files of a given campaign?   As fare as I remember the branches are uniform in the NanoAOD file for a given data/MC campaign. 

We can discuss more about this in a dedicated issue I agree.

PR #106, opened 2023-11-27
Improving plotting scripts

andreypz:
There are three main changes in this PR:
1. Simplify arguments for the `make_plots.py` script. By default one only requires to provide the directory of the output from the runner. Coffea file and params are deducted from there. Output location of the plots is then `<dir>/plots/`. One could overwrite those using the same arguments as before.
2. Fix the `sample_groups` functionality. It is broken in maser. (isMC check is needed for the merged dataset name).
3. A new option `exclude_samples` is added in the plotting config.


andreypz:
Also added:
4. An option for rescale certain samples.

Ready for review.

valsdav (approved):

scripts/plot/make_plots.py
```diff
@@ -10,30 +10,44 @@
 from pocket_coffea.parameters import defaults
 
 parser = argparse.ArgumentParser(description='Plot histograms from coffea file')
-parser.add_argument('--cfg', help='YAML file with all the analysis parameters', required=True)
+parser.add_argument('input_dir', help='Directory with cofea files and parameters', type=str)
```
mmarchegiani, line 13: Shouldn't it be `--input_dir` instead of `input_dir`?
Also, now the `make_plots.py` script is crashing if the `--input_dir` parameter is not passed as an argument. I think it would be desirable to be able to run the script by passing the input dir (`--input_dir`) or the specific files (with `--cfg`, `-i`, `-o`)

andreypz: Well, it was intentional. A positional *required* argument don't need the `--prefix`.  
And it should crash if the argument not provided.

> it would be desirable to be able to run the script both by passing the input dir (--input_dir) or the specific files (with --cfg, -i, -o)
Ok, I understand, but I imagine most users don't want that. Default arguments work and require less typing. Advanced users can add additional arguments while still providing the input dir. In fact I am not sure what is the use case when all arguments need to be non-default?

PR #112, opened 2023-11-30
Fix crash due to missing rescale_sample parameter in plotting

andreypz:
Fixing a bug introduced in my previous commit:
when the `rescale_samples` is not present in the config, plotting script was crashing. This is fixed here

### DIFF
--- a/pocket_coffea/utils/plot_utils.py
+++ b/pocket_coffea/utils/plot_utils.py
@@ -41,6 +41,9 @@ class Style:
         self.has_exclude_samples=False
         if "exclude_samples" in style_cfg:
             self.has_exclude_samples = True
+        self.has_rescale_samples=False
+        if "rescale_samples" in style_cfg:
+            self.has_rescale_samples = True
         self.set_defaults()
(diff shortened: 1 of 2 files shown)


mmarchegiani:
The fix to avoid crashing when `rescale_samples` is not in the plotting config looks good.
About the `exclude_samples` key introduced in a previous PR, I think that it is currently not working since the samples are excluded only when `self.style.has_samples_groups` is False (https://github.com/PocketCoffea/PocketCoffea/blob/53d5b7585710e49fa75917b737790140dc8be8fb/pocket_coffea/utils/plot_utils.py#L399).

The desirable behaviour would be to have the histograms excluded even if there is no `samples_groups` key in the plotting config.
I propose to define a new method `exclude_samples()`, that is called in the constructor of the `Shape` class after `self.group_samples()` to handle this single operation.

Could you please fix that as well?
Thank you!

andreypz:
Indeed, it was not implemented correctly. Now should be fixed in my last commit.

PR #118, opened 2023-12-19
Adding jet C-tagging, improving Parsl runner, logger, and minor bugfixes

andreypz:
Here is a description of the changes in the order of appearance:
* `jets.py`: allow for jets sorting based on CvsL score
* `histograms.py`: add CvsL and CvsB scores in the default list of histograms for jets
   * Also adding a counter on the axis label to distinguish distributions when saving multiple objects (jet 1, 2, 3 etc).
* `dataset.py`: just adding one more check for valid dataset names I stumbled upon. And raising Exception
*  `plot_utils.py`: if a sample specified in the `group_samples` in the config does not exist - the plotting would crash. This fix avoids the crash.
* `rucio.py`: printing a warning that `.sites_map.json` already exists. I encountered an issue that one T2 site has changed their xrootd prefix and so code crashed.  It took me a while to figure out that the reason was that `.sites_map.json` was outdated. Hence this warning. Perhaps a better solution would be to avoid saving any such *hidden* file. Just creating this map on the fly would be better, in my opinion. It does not take much runtime.
* `make_plots.py`: only some comments
* `runner.py`:  a few changes here:
   *  The `logfile.log` is to be placed in the `outputdir`. However, for backward compatibility, this file is also copied to `last_run.log`
  * Changing the Console logger level when running with Parsl. This is because Parsl also writes its logs with logger and they are too much to be shown on stdout. They are still saved in the logfile.
  * Running with Parsl is now done per-dataset by default (same as with Dask), with a `--full` option to run all datasets at once.
* `test_package.py`: just make use of this script to print version. I think the`tests` directory could be used for such simple test scripts. But currently the scripts there don't look very useful

### DIFF
--- a/pocket_coffea/lib/jets.py
+++ b/pocket_coffea/lib/jets.py
@@ -234,6 +234,8 @@ def btagging(Jet, btag):
     return Jet[Jet[btag["btagging_algorithm"]] > btag["btagging_WP"]]
 
 
+def CvsLsorted(jets, ctag):
+    return jets[ak.argsort(jets[ctag["tagger"]], axis=1, ascending=False)]
 
 
 def get_dijet(jets):
--- a/pocket_coffea/parameters/histograms.py
+++ b/pocket_coffea/parameters/histograms.py
@@ -150,6 +150,22 @@ default_axis_settings = {
         "lim": (0, 1),
         'label': "AK4 DeepJet b-tag score",
     },
+    'jet_btagDeepFlavCvL': {
+        "field": "btagDeepFlavCvL",
+        "bins": 50,
+        "start": 0.0,
+        'stop': 1.0,
+        "lim": (0, 1),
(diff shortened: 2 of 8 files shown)


andreypz:
Please consider this PR, it is ready for review.

valsdav:
>     * `rucio.py`: printing a warning that `.sites_map.json` already exists. I encountered an issue that one T2 site has changed their xrootd prefix and so code crashed.  It took me a while to figure out that the reason was that `.sites_map.json` was outdated. Hence this warning. Perhaps a better solution would be to avoid saving any such _hidden_ file. Just creating this map on the fly would be better, in my opinion. It does not take much runtime.

Sure I will open an issue to track this. In the past I thought about keeping the file just 24 hours.

valsdav (approved):

PR #121, opened 2024-01-16
feat: Improved executors interface and generalized runner script

valsdav:
This PR aims to improve the executor configuration for the `runner.py` script. 

Until now different executor have been configured in an hard coded way in the runner script itself. 

After this change the executors will be setup in separate factory modules.  The core code will provide setup for specific sites (lxplus, T3s, analyses facilities...), and the user will be able to provide custom executor setup scripts directly to the runner. 

This should improve and generalize the runner interface.


config/tests/custom_run_options.yaml
```diff
@@ -0,0 +1,5 @@
+chunksize: 10000
+scalout: 10
```
mmarchegiani, line 2: This should be `scaleout`, right?

pocket_coffea/parameters/executor_options_defaults.yaml
```diff
@@ -0,0 +1,36 @@
+general: 
+  scalout: 1
```
mmarchegiani, line 2: `scalout` -> `scaleout`

pocket_coffea/executors/executors_T3_PSI_CH.py
```diff
@@ -0,0 +1,116 @@
...
+import socket
+from coffea import processor as coffea_processor
+from .executors_base import ExecutorFactoryABC
+from .executors_base import IterativeExecutorFactory, FuturesExecutorFactory
+from pocket_coffea.utils.network import check_port
+from pocket_coffea.parameters.dask_env import setup_dask
+
+    
+
+class DaskExecutorFactory(BaseExecutorABC):
```
mmarchegiani, line 12: Shouldn't it be:
`class DaskExecutorFactory(ExecutorFactoryABC):` ?

`BaseExecutorABC` is not defined elsewhere.

PR #125, opened 2024-01-25
Make it run on 2022 and 2023 datasets; and sign(genWeight) scaling 

andreypz:
As the title suggests, this PR is for enabling 2022 and 2023 runs.
Both years are split in two periods with the following `year` parameters to distinguish them:
`2022_preEE`, `2022_postEE`, `2023_preBPix`, `2023_postBPix`

In addition to that, there are incompatibilities of the NanoAOD variables between Run3 and Run2. 
In particular, `Jet.puID` does not exist. To avoid the crash I create a new parameter, `params['run_period']='Run3'` which is used to distinguish a Run3 processor. 
There is also incompatible Electron id variable: `id: mvaFall17V2Iso_WP80`  for Run2 and  `id: mvaIso_WP80`  for Run3. I did not think of anything better than to change it manually when running over Run3. But this should be dealt with in a better way somehow.

Other changes in this PR are some improvements for the old runner script. 

The PR is ready for review.


andreypz:
Pushed another commit to solve the #130

valsdav:
Thanks @andreypz having a look!

andreypz:
Yes, we will test the new runner and implement the modifiers needed to run at RWTH Aachen cluster. One of our new student is looking into this.

pocket_coffea/lib/jets.py
```diff
@@ -216,10 +216,13 @@ def jet_selection(events, jet_type, params, leptons_collection=""):
         dR_jets_lep = jets.metric_table(events[leptons_collection])
         mask_lepton_cleaning = ak.prod(dR_jets_lep > cuts["dr_lepton"], axis=2) == 1
 
-    if jet_type == "Jet":
+    if 'run_period' not in params or params['run_period']!='Run3': # puid does not exist for Run3 samples
```
valsdav, line 219: Let's discuss a moment about this approach. In my opinion the run_period is a charactheristic of the dataset, not of the parameters.  We should follow the same pattern as the `year` for many corrections, adding the `run_period` metadata key to the dataset or automatically assigning it from the `year` field.

This info is accessing from `events.metadata` and we can preload it in the `load_metadata()` function of the base processor like the year  https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/workflows/base.py#L121

pocket_coffea/workflows/base.py
```diff
@@ -806,11 +808,22 @@ def process(self, events: ak.Array):
...
 
-    def rescale_sumgenweights(self, sumgenw_dict, output):
+    def rescale_sumgenweights(self, output):
         # rescale each variable
+
         for var, vardata in output["variables"].items():
             for sample, dataset_in_sample in vardata.items():
                 for dataset, histo in dataset_in_sample.items():
+                    #print("Scaling", var, sample, dataset)
+                    #print(self.cfg.weights_config)
```
valsdav, line 818: please remove commented debugging printout

valsdav (changes requested):
Thanks a lot for all the work. I think the only think to discuss is the comment  about the run_period specialization.

pocket_coffea/lib/jets.py
andreypz: I agree with this. Just did not find how to get the info before. The events.metadata works, so switching to this.

I also agree that it would be good to have a single `run_period` attribute of a dataset in metadata, determined from `year`. But  I did not implement this now.

pocket_coffea/workflows/base.py
andreypz: Done

pocket_coffea/lib/jets.py
valsdav: Ok fine, I think the new implementation is fine and we can then elaborate more on how to handle the run_period automatically. Probably a new default parameters set will do.

PR #135, opened 2024-02-21
Add coffea.casa executor

oshadura:
Based on https://pocketcoffea.readthedocs.io/en/main/running.html#adding-support-for-a-new-executor-site

### DIFF
--- a/pocket_coffea/executors/executors_casa.py
+++ b/pocket_coffea/executors/executors_casa.py
@@ -1,0 +1,42 @@
+from pocket_coffea.executors.executors_base import ExecutorFactoryABC
+from pocket_coffea.executors.executors_base import IterativeExecutorFactory, FuturesExecutorFactory
+from coffea import processor as coffea_processor
+    
+
+class DaskExecutorFactory(ExecutorFactoryABC):
+
+    def __init__(self, run_options, outputdir, **kwargs):
+        self.outputdir = outputdir
+        super().__init__(run_options, **kwargs)
+        
+    def setup(self):
+        ''' Start the DASK cluster here'''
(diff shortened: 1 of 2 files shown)


valsdav:
Hi @oshadura thanks a lot!  
Yesterday we moved the scripts inside `pocket_coffea/scripts` to implement a CLI based on click. 
Can you please fix the conflict? :)

pocket_coffea/executors/executors_casa.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self, run_options, outputdir, **kwargs):
+        self.outputdir = outputdir
+        super().__init__(run_options, **kwargs)
+        
+    def setup(self):
+        ''' Start the DASK cluster here'''
+        # At coffea-casa we have preconfigured Dask HTCondor cluster for you, please just use it available at tls://localhost:8786
+        from distributed import Client
+        self.dask_client = Client("tls://localhost:8786")
```
valsdav, line 16: If this port always the same? Is it needed to put it in the default parameters?

pocket_coffea/executors/executors_casa.py
```diff
@@ -0,0 +1,42 @@
...
+        # Install pocket_coffea to dask workers via Dask scheduler plugin PipInstall
+        # PLEASE CHANGE if you are using fork or different tag / branch
+        from dask.distributed import PipInstall
+        plugin = PipInstall(packages=["git+https://github.com/PocketCoffea/PocketCoffea.git@main"])
```
valsdav, line 20: Can be configured later

pocket_coffea/executors/executors_casa.py
oshadura: It is always the same for each user

PR #142, opened 2024-03-01
JET PU Id for 2016 and naf executor

phnattla:
Added JET PU Id wps for 2016 and merged naf executor written by @valsdav


pocket_coffea/lib/jets.py
```diff
@@ -217,9 +217,28 @@ def jet_selection(events, jet_type, params, leptons_collection=""):
         mask_lepton_cleaning = ak.prod(dR_jets_lep > cuts["dr_lepton"], axis=2) == 1
 
     if jet_type == "Jet":
-        # Selection on PUid. Only available in Run2 UL, thus we need to determine which sample we run over:
-        if events.metadata["year"] in ['2016_PreVFP', '2016_PostVFP','2017','2018']:
-            mask_jetpuid = (jets.puId >= cuts["puId"]["value"]) | (
+        # Selection on PUid. Only available in Run2 UL, thus we need to determine which sample we run over;
+        # Due to bug in JetPUid production, values are different for 2016 
+        # (see here https://twiki.cern.ch/twiki/bin/view/CMS/PileupJetIDUL#Trainings)
+        jetPuID_values = {
```
valsdav, line 223: This map should become part of the parameters instead of being hardcoded in the function. 

The location should be here https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/parameters/jet_scale_factors.yaml#L21,  adding a new subkey called "JetPuID_values". Then they can be used here :)

valsdav (changes requested):
Thanks a lot @phnattla!

pocket_coffea/lib/jets.py
phnattla: Good point, I changed the Implementation accordingly.

pocket_coffea/parameters/jet_scale_factors.yaml
```diff
@@ -33,3 +33,23 @@ jet_scale_factors:
         file: /cvmfs/cms.cern.ch/rsync/cms-nanoAOD/jsonpog-integration/POG/JME/2018_UL/jmar.json.gz
         name: PUJetID_eff
 
+  jet_puId_WPs:
+      # Jet PU ID working points for different periods, due to bug values are different for 2016 
+      # (see here https://twiki.cern.ch/twiki/bin/view/CMS/PileupJetIDUL#Trainings)
```
valsdav, line 38: Looking at it we should do something like : 

```
jet_puId:
       2016_PreVFP:
           file:  xxxxxxxxxxxxxxxxxxxxxxxxx
           working_point: 
                  L : 1
                  M: 3
                  T: 7
```

What do you think? The function should be changed accordingly but should be straight forward.

PR #144, opened 2024-03-04
Improved b-tagging parameters

mmarchegiani:
The scheme for b-tagging parameters is updated to support multiple working points of b-tagging:
- The b-tagging working points for Run-2 have been updated to the latest recommendation from the BTV-POG (https://btv-wiki.docs.cern.ch/PerformanceCalibration/#working-points).
- The Run-3 parameters are updated according to the latest recommendation for the DeepJet tagger. If another tagger has to be used, the user needs to define their custom tagger and working points parameters to override the defaults.
- A key `reference` is included in the b-tagging parameters in `btagging.yaml` to trace the source of the parameters. This entry needs to be updated whenever the WPs are updated and helps the user to keep track of the origin of the defaults.
- The function to apply b-tagging in the processor and the helper functions to require a certain number of b-tagged jets have been updated accordingly, taking now the working point as an additional argument.
- The `wp` argument is set to `M` by default, meaning that the medium working point is applied if the argument is not specified.
- A small fix to the base processor is performed in order to dump arrays to parquet files.

### DIFF
--- a/pocket_coffea/workflows/tthbb_base_processor.py
+++ b/pocket_coffea/workflows/tthbb_base_processor.py
@@ -44,7 +44,7 @@ class ttHbbBaseProcessor(BaseProcessorABC):
             self.events, "Jet", self.params, "LeptonGood"
         )
         self.events["BJetGood"] = btagging(
-            self.events["JetGood"], self.params.btagging.working_point[self._year], wp="M"
+            self.events["JetGood"], self.params.btagging.working_point[self._year], wp=self.params.object_preselection.Jet["btag"]["wp"]
         )
 
         # self.events["FatJetGood"], self.jetGoodMask = jet_selection(
--- a/pocket_coffea/workflows/tthbb_base_processor.py
+++ b/pocket_coffea/workflows/tthbb_base_processor.py
@@ -44,7 +44,7 @@ class ttHbbBaseProcessor(BaseProcessorABC):
             self.events, "Jet", self.params, "LeptonGood"
         )
         self.events["BJetGood"] = btagging(
-            self.events["JetGood"], self.params.btagging.working_point[self._year]
+            self.events["JetGood"], self.params.btagging.working_point[self._year], wp="M"
         )
 
         # self.events["FatJetGood"], self.jetGoodMask = jet_selection(
--- a/pocket_coffea/lib/columns_manager.py
+++ b/pocket_coffea/lib/columns_manager.py
@@ -8,7 +8,7 @@ import awkward as ak
 class ColOut:
     collection: str  # Collection
     columns: List[str]  # list of columns to export
-    flatten: bool = True  # Flatten by defaul
+    flatten: bool = True  # Flatten by default
     store_size: bool = True
     fill_none: bool = True
     fill_value: float = -999.0  # by default the None elements are filled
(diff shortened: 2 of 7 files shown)


mmarchegiani:
FYI @andreypz
We should also update the Run-3 parameters including the L, M and H working points, to make this PR work for Run-3. Maybe we can double-check the parameters before merging.

pocket_coffea/workflows/tthbb_base_processor.py
```diff
@@ -44,7 +44,7 @@ def apply_object_preselection(self, variation):
             self.events, "Jet", self.params, "LeptonGood"
         )
         self.events["BJetGood"] = btagging(
-            self.events["JetGood"], self.params.btagging.working_point[self._year]
+            self.events["JetGood"], self.params.btagging.working_point[self._year], wp="M"
```
valsdav, line 47: I would keep the chosen working point as part of the configuration of the analysis and not as part of the processor.  
This may fix the `object_preselection` part of the params within the jet dictionary

mmarchegiani: Agree, maybe we can do it by passing the b-tag working point in the `workflow_options` in the Configurator?
If not passed, we can keep the medium working point as default.

valsdav: In my opinion it is not a workflow option, but a preselection params like others, so to be put here: https://github.com/PocketCoffea/AnalysisConfigs/blob/main/configs/ttHbb/semileptonic/ttbar_background/params/object_preselection_semileptonic.yaml for example.

valsdav:
We may also remove the `tthbb_base_processor.py` from the central repository.

pocket_coffea/workflows/tthbb_base_processor.py
mmarchegiani: Makes sense. Should we add an entry to the `Jet` key to specify the btag WP?
I was thinking something like:
```
Jet:
    ...
    btag:
        wp: M
```
and this key is used to define the collection `BJetGood` in the object preselection of custom processors.

Instead, the b-tagging algorithm, the working points and SF can stay in the central parameters in https://github.com/PocketCoffea/PocketCoffea/blob/337ef90fbba299be4a2c75586a4808fd9c1ed950/pocket_coffea/parameters/btagging.yaml

valsdav: sure. The compatibility of the configuration for the preselection and the function is there if one follows the default structure.   

If a user needs more complex selections or configuration both the function and the parameters can be customized. I think this is a sensible default

mmarchegiani: Update: now the b-tagging WP is read from the object preselection parameters [here](https://github.com/PocketCoffea/PocketCoffea/blob/5fa68398e5c02892ea9f1e67fd0a2a395a974ac2/pocket_coffea/workflows/tthbb_base_processor.py#L47).
However the structure of the dictionary `self.params.object_preselection.Jet["btag"]` is not enforced anywhere and it's up to the user how to pass the b-tagging WP.

mmarchegiani:
The Run-3 WP parameters have been updated as well. From my side, this PR is ready for merging.
FYI @valsdav

PR #154, opened 2024-04-15
Run3 sf defaults additions

phnattla:
Added some new defaults for Run3:

1.  PUid: Added 2022/2023 periods in `pocket_coffea/parameters/pileup.yaml`
2. Lumi: Corrected the defaults in `pocket_coffea/parameters/lumi.yaml` according to the current recommentations (with normtag, see [here](https://twiki.cern.ch/twiki/bin/view/CMS/LumiRecommendationsRun3))
3. Added feature, that iso cut on electrons is not needed to be given in params, since current recommendation includes no iso cut on electrons, since mva ID includes iso information
4. sf_ele_reco/id: 
  - Changed mapping of params for electron id/reco sfs, since pt_regions are different for Run3 and therefore depend on the year
 - Adapted `sf_ele_reco` function to account for the changes and enable more than 2 pt_regions
 - Cross checked implementation for 2018, seems to give same results, still left old function in for reference
5. Added "jetVetoMaps" params according to current recommentdation, could also add my corresponding function evaluating the correctionlib file to `cut_functions.py`? Currently have it in my `custom_cut_functions.py`

### DIFF
--- a/pocket_coffea/lib/scale_factors.py
+++ b/pocket_coffea/lib/scale_factors.py
@@ -19,9 +19,10 @@ def get_ele_sf(
         map_name = electronSF.JSONfiles[year]["name"]
 
         if key == 'reco':
-            sfname = electronSF["reco"][pt_region]
+            sfname = electronSF.JSONfiles[year]["reco"][pt_region]
         elif key == 'id':
-            sfname = electronSF["id"]
+            print(params.object_preselection["Electron"]["id"])
+            sfname = electronSF["id"][params.object_preselection["Electron"]["id"]]
 
         # translate the `year` key into the corresponding key in the correction file provided by the EGM-POG
         year_pog = electronSF["era_mapping"][year]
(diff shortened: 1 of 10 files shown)


pocket_coffea/lib/scale_factors.py
```diff
@@ -19,9 +19,10 @@ def get_ele_sf(
         map_name = electronSF.JSONfiles[year]["name"]
 
         if key == 'reco':
-            sfname = electronSF["reco"][pt_region]
+            sfname = electronSF.JSONfiles[year]["reco"][pt_region]
         elif key == 'id':
-            sfname = electronSF["id"]
+            print(params.object_preselection["Electron"]["id"])
```
valsdav, line 24: Please remove debug printouts

pocket_coffea/lib/scale_factors.py
```diff
@@ -106,6 +118,61 @@ def get_mu_sf(params, year, pt, eta, counts, key=''):
...
+    sfdown_reco = ak.prod(
+        ak.concatenate(sfdown_reco, axis=1), axis=1
+    )
+
+    return sf_reco, sfup_reco, sfdown_reco
+
+
+
+
+def sf_ele_reco_old(params, events, year):
```
valsdav, line 175: It is better to remove "old" implementations if updated ones have been added.

pocket_coffea/parameters/executor_options_defaults.yaml
```diff
@@ -44,6 +44,7 @@ parsl-condor@DESY_NAF:
   logs-dir: logs_parsl
   queue: ""
   walltime: "12:00:00"
+  ignore-grid-certificate: false
```
valsdav, line 47: this has been now added to the general defaults.  But it is not harmful to keep it also in the defaults by-sites if you wish

pocket_coffea/parameters/lepton_scale_factors.yaml
```diff
@@ -2,28 +2,59 @@
 lepton_scale_factors:
   ####################################################################################################
   electron_sf:
-    reco:
-      pt_gt_20: RecoAbove20
-      pt_lt_20: RecoBelow20
 
-    id: wp80iso
+    id:
```
valsdav, line 6: For documentation reasons, it would be nice to add here a comment to explicit that the map here is linking the SFname for Iso and the label that should be used in the object preselections.  

Just to refresh my memory: why don't we use the same label for the object preselection to avoid this map?

valsdav (changes requested):
Hi @phnattla thanks a lot for the PR. Looks good!

I agree it would be nice to also include your jet veto implementation in the provided cuts_functions in `pocket_coffea.lib`. Can you add it in this PR please? Thanks!

pocket_coffea/parameters/lepton_scale_factors.yaml
phnattla: The label for "id", that is used in the object preselection is the name of the corresponding field in NanoAOD (see e.g. [here](https://cms-nanoaod-integration.web.cern.ch/autoDoc/NanoAODv12/2022/2023/doc_EGamma_Run2022F-22Sep2023-v1.html#Electron) . This id requirement is also checked directly without further mapping [here](https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/lib/leptons.py#L21). Therefore, if one were to use the "sf labels" in the object preselection, one would have to introduce a mapping to the corresponding NanoAOD field, which is i think less intuitive.

valsdav:
Thanks @phnattla !

PR #160, opened 2024-04-22
lib: leptons, jets, limit exports to functions

matintorkian:
Since object.py is using `import *`, It is best practice to limit the exports to defined functions explicitly. For more, see:

https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

Clean up unused `importlib`.

## P.S:

I tried following the [guidelines](https://github.com/PocketCoffea/PocketCoffea/blob/97771bf4af4dc92abbfa630090fcddce250cf829/.github/CONTRIBUTING.md), however running `nox -s lint` raised an error complaining that the poetry configuration is invalid:

>           RuntimeError: The Poetry configuration is invalid:
>           - data.extras.pipfile_deprecated_finder[2] must match pattern ^[a-zA-Z-_.0-9]+$

I would appreciate if you could point me in the right direction, and if I missed something.

I have already set up a `.venv` directory and followed these instructions:

```bash
python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install -v -e '.[dev]' # needed the single quotations since zsh thought it is a regex
```
However, `nox` still cannot successfully complete its `test` and other operations.

I am running python version 3.12 on macOS Sonoma `Version 14.4 (23E214)`.


valsdav:
Thanks a lot. I agree on the import practice. 

About the contribution guidelines: we need to cleanup the testing/linting chain.  Any help is very appreciated also on that front :)

valsdav:
Hi @matintorkian is this PR still needed?

PR #164, opened 2024-05-06
Update Mu SFs for latest convention from POG

andreypz:
Simple change to make Muon SFs work again. The POG unified the corrections for Run2 and Run3, now year/era is not needed anymore. It is defined by the correction file itself.


valsdav:
Thanks for the quick fix!

PR #171, opened 2024-05-22
VHcc analysis. Adding variables.

ValVau:
Adding new variables important for VHcc analysis.
+BTV Commissioning. Soft lepton selection

### DIFF
--- a/pocket_coffea/lib/leptons.py
+++ b/pocket_coffea/lib/leptons.py
@@ -17,7 +17,9 @@ def lepton_selection(events, lepton_flavour, params):
         # Requirements on SuperCluster eta, isolation and id
         etaSC = abs(leptons.deltaEtaSC + leptons.eta)
         passes_SC = np.invert((etaSC >= 1.4442) & (etaSC <= 1.5660))
-        passes_iso = leptons.pfRelIso03_all < cuts["iso"]
+        passes_iso = True
+        if "iso" in cuts.keys():
+            passes_iso = leptons.pfRelIso03_all < cuts["iso"]
         passes_id = leptons[cuts['id']] == True
 
         good_leptons = passes_eta & passes_pt & passes_SC & passes_iso & passes_id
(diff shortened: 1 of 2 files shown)


pocket_coffea/lib/leptons.py
```diff
@@ -17,9 +17,7 @@ def lepton_selection(events, lepton_flavour, params):
         # Requirements on SuperCluster eta, isolation and id
         etaSC = abs(leptons.deltaEtaSC + leptons.eta)
         passes_SC = np.invert((etaSC >= 1.4442) & (etaSC <= 1.5660))
-        passes_iso = True
-        if "iso" in cuts.keys():
-            passes_iso = leptons.pfRelIso03_all < cuts["iso"]
+        passes_iso = leptons.pfRelIso03_all < cuts["iso"]
```
valsdav, line 20: Hi! I just have a question about this change.  Not in all the analysis this "iso" key is used, that's why there was this if clause. Does that cause problems to you ? 

We need to keep the function general enough to support different data-taking periods configuration. Alternatively we can create different functions

ValVau: Actually not, we can also keep general and stay with an if cause

valsdav: Ehi @ValVau if you push a commit to keep the if I think the PR is good to go!

ValVau: Grazie mille. I pushed a commit keeping the if. Please take a look.

pocket_coffea/lib/leptons.py
```diff
@@ -33,6 +33,35 @@ def lepton_selection(events, lepton_flavour, params):
 
     return leptons[good_leptons]
 
+def soft_lepton_selection(events, lepton_flavour, params):
```
valsdav, line 36: Another comment on this. 

Looking at the code it seems that this function differs from `lepton_selection` only because of a dxy cut. 
Would it be possible to add a `if  "dxy" in cuts` check also for that and keep a single function?

ValVau: This function has an inverted iso criterium as well, which is a required option, apart from dxy. Would you want to integrate it in the main function?

valsdav: Right! I read too quick. In this case it is fine to keep it separate, it has a different scope

PR #173, opened 2024-05-24
change run option

felixzinn:
`--full` renamed to `--process-separately`
docs updated
new name better reflects its purpose
it is clear that the `--full` (now --process-separately) and `--test` option are not related


valsdav:
Thanks @felixzinn! I think this is good to go!

PR #174, opened 2024-05-27
Implementing JECs for 2022 and 2023 eras

andreypz:
JEC and JER corrections for 2022 and 2023 implemented


andreypz:
Also added Electron SFs for 2023, which were missing before.

This is ready for review @valsdav

pocket_coffea/lib/scale_factors.py
```diff
@@ -25,16 +25,35 @@ def get_ele_sf(
...
-            year_pog, "sfup", sfname, eta.to_numpy(), pt.to_numpy()
-        )
-        sfdown = electron_correctionset[map_name].evaluate(
-            year_pog, "sfdown", sfname, eta.to_numpy(), pt.to_numpy()
-        )
+        
+        if year in ["2023_preBPix", "2023_postBPix"]:
+            # Starting from 2023 SFs require the phi:
+            if key == 'reco' and pt_region == 'pt_lt_20':
+                # It also appears that for RecoBelow20 SFs the eta must be positive (absolute value).
```
valsdav, line 32: and not for RecoAbove20? Probably they didn't have enough event and folded the histogram in eta?

Do you think we should cross check with egamma if this is a bug?

andreypz: I have already y contacted Riccardo, who pushed the correction to jsonpog gitlab. He told me that this is the case only for Below20 corrections. Indeed due to low stats.
He also said that maybe there will be a new version that would work with negative etas (just making SFs symmetric over). But for now we need this workaround.

Issue #177, opened 2024-06-05
Jet Enery Corrections are not applied to Data (but they should be)


andreypz:
In the way JECs are implemented in PocketCoffea, they are only applied when `shape_variations` in the config are specified.
Moreover, those variations only seem to be applied to MC and [not to Data](https://github.com/PocketCoffea/PocketCoffea/blob/f7342feefd55697c5dc6551e9cd84726519841af/pocket_coffea/workflows/base.py#L602-L604). This makes sence for most correction but not JECs. The JECs should be applied to Data as well, [Ref](https://cms-jerc.web.cern.ch/JEC/).

For Run2 samples this is a minor issue since jets are already corrected in NanoAOD, but for Run3 this is a big issue, since in NanoAOD they jets seem to be way off the scale.
I haven't come up with a good way to implement the correction, thus making this issue for now.

valsdav:
Thanks for the bringing this up @andreypz. Indeed for Run2 UL everything was applied, but now for Run3 we need to calibrate them.  Let me have a look at the code and we can discuss how to do it.

valsdav:
Solved by #182

PR #182, opened 2024-06-17
Adding JEC for Run3 and apply JES on data

valsdav:
For Run3 datasets, we need to apply jet energy scale corrections also on nominal MC and on Data. 

- Refactored `jets_calibration.yaml` parameters to contain both MC and Data JES files 
- Refactored `base` workflow to apply JES on nominal and not only if JES variations are requested

Addressing issue #177


andreypz:
I run with this branch and it works for me!
I also confirm that the Jet's scale is much improved in Data (only looked at 2023 ). I would say it can be merged.

One thing is not clear though is if JER smearing is applied to MC, or not. Looking at the pt(jj)/pt(ll) ratio, the MC is a bit narrower than Data, which suggests the the jet's pT are not smeared (enough).

dilep_dijet_ratio_2023_postBPix_baseline_2L2J_no_ctag

valsdav:
Ok nice! I see that JER is included for Run3 MC in the configuration. 
If it is defined as part of the correction files it should be applied alongside the JEC. 
Maybe you can define another correction set without JER (https://github.com/PocketCoffea/PocketCoffea/blob/jec_nominal/pocket_coffea/parameters/jets_calibration.yaml#L215) in your custom parameters and see if we see a difference without the JER applied?

valsdav:
I also noticed this warning when preparing the jet corrector maps:

```
Creating AK4PFPuppi jet calibrator for 2022_postEE
/t3home/dvalsecc/micromamba/envs/pocket-coffea/lib/python3.9/site-packages/coffea/lookup_tools/txt_converters.py:166: UserWarning: binning for file for Summer22EEPrompt22_JRV1_MC_PtResolution_AK4PFPuppi is malformed in variable JetEta
```

andreypz:
I commented out the `MC_PtResolution_AK4PFPuppi.jr.txt.gz` and `MC_SF_AK4PFPuppi.jersf.txt.gz` files from the jecs config and rerun.
I do see an effect, here is a plot:
dilep_dijet_ratio_2023_postBPix_baseline_2L2J_no_ctag

The MC distribution does look a little bit narrower, but the difference is not much with respect to JER corrected plot I posted before...

I also get the Warnings about the JER correction files, but I have not checked 2022 plots yet.

But I think this PR itself is good to be merged. 
The issues need to be discussed with JETMET group (maybe after switching to correctionlib implementation).

valsdav:
Ok! Merging now the technical PR. We will followup on the actual correction themselves. 

For the correctionlib implementation. We will discuss with central coffea to understand how to integrate correctionlib in the CorrectedJetFactory that we are using right now :+1:. We don't want an in-house solution for that.

PR #187, opened 2024-06-21
Enhancement/plotting

### LINKED ISSUE #123: Save plots in pdf and svg format
Currently only the `.png` output format is supported for plots.
Allow the users to save plots in the `.pdf` and `.svg` format through an optional argument.

felixzinn:
- the default color scheme for the stacked histograms is now taken from the CMS guidelines (https://cms-analysis.docs.cern.ch/guidelines/plotting/colors/) to be color-vision-deficiency-friendly
- the make_plots.py script and the corresponding plotting function now have an option to specify the file format (possibly closes #123 )


pocket_coffea/utils/plot_utils.py
```diff
@@ -513,9 +514,10 @@ def _get_stacks(self, cat, spliteras=False):
                 self.nevents = dict(
                     sorted(self.nevents.items(), key=lambda x: x[1], reverse=reverse)
                 )
-                color = iter(cm.gist_rainbow(np.linspace(0, 1, len(self.nevents.keys()))))
-                # Assign random colors to each sample
-                self.colors = [next(color) for d in self.nevents.keys()]
+                # create cycler from the colormap and instantiate it to get iterator
+                color = cycler("color", hep.styles.cms.cmap_petroff)()
+                # Assign colors from cycle to samples
```
valsdav, line 519: I assume that if no colors are explicitely assigned, we want to take them from the cycler. But is this keeping the same mapping with the samples for all the plots?

felixzinn: its not cycling across different shape objects, if thats what you mean
each time the cycler is instantiated, its a new cycler starting from the first element in the list which is taken from `hep.styles.cms.cmap_petroff` 
so the first sample in the sorted dict `self.nevents` is always assigned the first color in the `cmap_petroff` list

valsdav: ok perfect, the colours will be consistent between plots.

valsdav:
Thanks a lot @felixzinn !

PR #191, opened 2024-06-25
Removed application of sumgenweight rescaling by default

valsdav:
SumGenWeight is used for weight rescaling only if `genWeight` or `signOf_genWeight` is used in the `inclusive` weights configuration for the sample.


andreypz:
Then, maybe also worth to add an `elif` for the case when both are specified, and raise an error?

valsdav:
I prefer to raise a warning and not an error to not trash the full processing at the end of the chain.  We can add an expection in the Configurator to check for that.

Issue #104, opened 2023-11-23
Remove Primary Vertex requirements from base workflow skim


valsdav:
The `nPV_good > 0` cut is sensible, but not general. It should be added by the user in the configuration, not hidden in the default skim function. 

https://github.com/PocketCoffea/PocketCoffea/blob/a580989ad6509ec485f20cca6b0533b3ba2ad7b0/pocket_coffea/workflows/base.py#L165C9-L166C1

valsdav:
The skim will look like

```python 
from pocket_coffea.lib.cut_functions import  eventFlags, get_nPVgood, goldenJson


cfg = Configurator( .....    
    skim = [ eventFlags,
             get_nPVgood(1),
             goldenJson,
)
```

PR #193, opened 2024-06-26
Removed default skim cuts from base processor

valsdav:
Removed default skim cuts from base processor and defined corresponding cut objects in the `cut_functions` lib. 

- nPV_good > 0 selection
- goldenJson
- event flags

Solves #104


valsdav:
Sure I will add those in the example configs and configurations. 

Also we need to start a "common steps" section of the documentation.

valsdav:
Once this PR will be merged, I will tag a new release and circulate the changelog to warn users about the default skim changes.

PR #195, opened 2024-06-28
Separated init and loading in the Configurator

valsdav:
The configurator constructor was initializing the Configurator but also loading all the necessary data from the dataset configuration. 

This PR separates the construction and loading the the Configurator to allow users to load the Configurator object to look at options,metadata, parameters, without having to load the actual datasets to begin the processing. 

This allows law to inspect the Configurator before the dataset files are defined.

@felixzinn


felixzinn:
Some law tasks need access to the config parameters without loading, potentially non existent, dataset configuration files.
The separation allows exactly that without raising errors.
👍🏼

PR #199, opened 2024-07-03
fix check of genWeight in weights config

felixzinn:
wrong syntax for check of `"genWeight"` and `"signOf_genWeight"` in weights config. genWeights have not been applied to the histograms

old syntax 
```python
elif "genWeight" in wei and not "signOf_genWeight":
    sumgenw_dict = output["sum_genweights"]
```
the `not "signOf_genWeight"` always evaluates to `False`, so histograms are not rescaled

changed to
```python
elif "genWeight" in wei and "signOf_genWeight" not in wei:
    sumgenw_dict = output["sum_genweights"]
```


valsdav:
Thanks for catching this!

PR #201, opened 2024-07-04
udate plotting of systematic shifts

felixzinn:
The `SystUnc` class, which stores information about a single systematic uncertainty, has a `plot` method, to plot nominal as well as up and down variations on the same axes.
This PR updates this method. Furthermore the `Shape` and `PlotManager` class can now call this method to plot the variations for the categories and shape objects.
It is possible to plot the absolute variations only as well as the ratio of the shifts to nominal.

### DIFF
--- a/pocket_coffea/utils/plot_utils.py
+++ b/pocket_coffea/utils/plot_utils.py
@@ -188,6 +188,38 @@ class PlotManager:
             for shape in shape_names:
                 self.plot_datamc(shape, syst=syst, spliteras=spliteras, format=format)
 
+    
+    def plot_systematic_shifts(self, shape, format="png", ratio=True):
+        """Plots the systematic shifts for all the variations."""
+        if self.verbose > 0:
+            print("Plotting systematic shifts for:", shape.name)
+        if shape.dense_dim > 1:
+            print(
+                f"WARNING: cannot plot systematic shifts for histogram {shape.name}"
+                f"with dimension {shape.dense_dim}."
+            )
+            print("The method `plot_systematic_shifts` will be skipped.")
+            return
+        if shape.is_data_only:
+            raise Exception(
+                "The systematic shifts cannot be plotted if the histogram is Data only."
+            )
+        shape.plot_systematic_shifts_all(ratio=ratio, format=format)
+
+    def plot_systematic_shifts_all(self, format="png", ratio=True):
+        """Plots the systematic shifts for all the shape objects."""
+        if self.workers > 1:
+            with Pool(processes=self.workers) as pool:
+                # Parallel calls of plot_systematic_shifts() on different shape objects
+                pool.map(
+                    partial(self.plot_systematic_shifts, format=format, ratio=ratio),
+                    self.shape_objects.values(),
+                )
+                pool.close()
+        else:
+            for shape in self.shape_objects.values():
+                self.plot_systematic_shifts(shape, format=format, ratio=ratio)
+
 class Shape:
     '''This class handles the plotting of 1D data/MC histograms.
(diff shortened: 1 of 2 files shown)


pocket_coffea/utils/plot_utils.py
```diff
@@ -236,8 +268,8 @@ def load_attributes(self):
             print(self.h_dict)
             print("samples:", self.samples_mc)
         
-        self.is_mc_only = True if len(self.samples_data) == 0 else False
-        self.is_data_only = True if len(self.samples_mc) == 0 else False
+        self.is_mc_only = len(self.samples_data) == 0
+        self.is_data_only = len(self.samples_mc) == 0
```
felixzinn, line 272: `True if... else False` can be simplified to comparison

pocket_coffea/utils/plot_utils.py
```diff
@@ -419,7 +451,7 @@ def group_samples(self):
                         isMC = isMC_d
                         self.sample_is_MC[sample] = isMC
                     elif isMC != isMC_d:
-                        raise Exception(f"You are collapsing together data and MC histogram!")
+                        raise Exception("You are collapsing together data and MC histogram!")
```
felixzinn, line 454: there was nothing to format here, so i removed the `f` string

pocket_coffea/utils/plot_utils.py
```diff
@@ -1048,44 +1128,145 @@ def _get_err2(self, stacks):
...
+
+        # setup figure and corresponding axes
+        self.log = log
+        self.define_figure(ratio=ratio, toplabel=toplabel)
+        
+        # plot histograms
+        # https://stackoverflow.com/a/65935165
         self.ax.hist(
-            self.style.opts_axes["xcenters"],
+            self.bins[:-1],
```
felixzinn, line 1229: before the change the number of bins was set to the default of matplotlib (10 at the time of writing) and the entries in `self.style.opts_axes["xcenters"]` were passed to `numpy.histogram` with 10 bins, which is not what we want.
We want to fill each bin with exactly 1 entry that has a weight according to the histogram (nominal/up/down).
(https://matplotlib.org/3.9.0/api/_as_gen/matplotlib.pyplot.hist.html)

felixzinn:
i did not add any script to produce these plots yet.
It should be similar to the `make_plots.py` script.
```python
from pocket_coffea.utils.plot_utils import PlotManager

plot_manager = PlotManager(...)
plot_manager.plot_systematic_shifts_all(...)
```
Does it make sense to create an extra script or should we add an option to the existing script?

valsdav:
> i did not add any script to produce these plots yet. It should be similar to the `make_plots.py` script.
> 
> ```python
> from pocket_coffea.utils.plot_utils import PlotManager
> 
> plot_manager = PlotManager(...)
> plot_manager.plot_systematic_shifts_all(...)
> ```
> 
> Does it make sense to create an extra script or should we add an option to the existing script?

I think it makes sense to create an additional option to the `make-plots` script, something like `--systematics-shifts`.

felixzinn (commented):
i added two flags `--systematics-shifts` and `--no-systematics-ratio`
now the shifts can be plotted using
```sh
pocket-coffea make-plots -i output.coffea -o plots --systematics-shifts
```
or
```sh
pocket-coffea make-plots -i output.coffea -o plots --systematics-shifts --no-systematics-ratio
```
if one does not want to have the ratio plot of the shifts

valsdav:
Hi Felix, looks good to me! if you happy about it we can merge

felixzinn:
yes, we can merge!
When i have time i will update the documentation 👍🏼

PR #206, opened 2024-07-18
Enhancement/plotting: 10-color scheme added

felixzinn:
The official CMS 10-color scheme is added.
Depending on the number of processes to plot in the histogram either the 6- or 10-color scheme is chosen.
If there are more then 10 processes to plot, a warning is printed.


pocket_coffea/utils/plot_utils.py
```diff
@@ -507,7 +525,13 @@ def _get_stacks(self, cat, spliteras=False):
                     for d in self.samples_mc
                 }
                 # create cycler from the colormap and instantiate it to get iterator
-                color = cycler("color", hep.styles.cms.cmap_petroff)()
+                cmap = CMAP_6 if len(self.nevents) <= 6 else CMAP_10
+                if len(self.nevents) > 10:
```
felixzinn, line 529: If there are more than 10 processes to plot in the histogram there are colors assigned to multiple processes.
This is most likely not desired since you cannot tell them apart.
Is a warning enough in this case or should we handle this somehow differently?

valsdav:
Thanks @felixzinn !

Issue #175, opened 2024-06-04
Blinding data from some hists in plotting


andreypz:
It may be useful (or necessary) to blind the data in some histograms. It would be nice if such feauture existed in pocket coffea.
I think this could be implemented with the following syntax in the plotting config:
```yaml
  blind_hists:
      categories:
               - SigRegion1
               - SigRegion2
       histograms:
               - mjj: [100, 150]
               - BDT: [0.5, 1]
               
```

andreypz:
Should be resolved by #211

PR #211, opened 2024-07-30
Enable blinding data for certain histograms

andreypz:
This will enable to blind some data distributions in Signal regions.
This addresses issue #175 
Documentation is also updated.

Also adding missing default values in variations for 2022/2023 eras.


andreypz:
This is ready for review @valsdav

valsdav:
very nice, thanks for implementing it.

PR #212, opened 2024-07-31
update electron trigger scale factors

felixzinn:
This PR updates the evaluation of the electron trigger scale factors for the correctionlib files provided by the EGM POG for Run3

Im not yet familiar with the new weights handling in PocketCoffea, so the changes need to be checked. There might be an easier way to do it.
@valsdav 

the lepton_scale_factors.yaml file could look like this
```yaml
lepton_scale_factors:
  electron_sf:
    trigger_sf:
      2022_preEE:
        name: <key for correction set>
        file: <json file>
        path: <trigger path as listed by EGM POG>
```


valsdav:
The implementation is fine I think. 
I think that we can implement this default correction from EGM in the central function and move the ttHbb implementation of the SF on the side of the analysis group with a custom weight.

felixzinn:
so this commit [fd2b21e270763517b9bf5abc1b81db0a75432bb8](https://github.com/PocketCoffea/PocketCoffea/pull/212/commits/fd2b21e270763517b9bf5abc1b81db0a75432bb8) should not be necessary
and we only keep this [20a174fdf46a2ea1f935926fb7782f2b8dadf422](https://github.com/PocketCoffea/PocketCoffea/pull/212/commits/20a174fdf46a2ea1f935926fb7782f2b8dadf422)

is that correct?

valsdav:
I agree. 

Then we can remove the custom trigger SF that is called in the current function and make it a `sf_ele_trigger_tthbb` weight.

felixzinn (commented):
to get the EGM scale factors, the syntax would be:

```python
from pocket_coffea.lib.weights.common import weights_run3

weights = {
    "common": {
        "inclusive": [
            "genWeight",
            "lumi",
            "XS",
            "sf_ele_trigger_egm",
            ...
        ]
    }
}
```

mmarchegiani:
Hi @felixzinn, thank you for the PR.

> so this commit [fd2b21e270763517b9bf5abc1b81db0a75432bb8](https://github.com/PocketCoffea/PocketCoffea/pull/212/commits/fd2b21e270763517b9bf5abc1b81db0a75432bb8) should not be necessary and we only keep this [20a174fdf46a2ea1f935926fb7782f2b8dadf422](https://github.com/PocketCoffea/PocketCoffea/pull/212/commits/20a174fdf46a2ea1f935926fb7782f2b8dadf422)
> 
> is that correct?

The Run3 part implemented in the function `sf_ele_trigger_EGM` looks good to me, and I think it can become the new default for the electron trigger SF and be moved to the function above `get_ele_sf()` in the `elif key == 'trigger'` case.
At this point, I would just name the correction `sf_ele_trigger` instead of `sf_ele_trigger_egm` since it becomes the new default.
The old custom Run2 electron trigger SF that I implemented can be moved to [pocket_coffea/lib/weights/common/weights_run2_UL.py](https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/lib/weights/common/weights_run2_UL.py) since it's a custom weight for Run2.

mmarchegiani:
Also, another comment about the parameters.
I would include some defaults in `pocket_coffea/parameters/lepton_scale_factors.yaml`, since the name of the correctionlib correction is constant and the path of the file is the central POG file in /cvmfs:
```yaml
lepton_scale_factors:
  electron_sf:
    trigger_sf:
      2022_preEE:
        name: "Electron-HLT-SF"
        file: /cvmfs/cms.cern.ch/rsync/cms-nanoAOD/jsonpog-integration/POG/EGM/2022_Summer22/electronHlt.json.gz
```

Instead, I would not specify a default `path` since this depends on the muon definition in the analysis, so the user is forced to check this parameter and define it in the custom parameters.

felixzinn (commented):
I moved the custom implementation into `weights_run2_UL.py` which means that `weights_classes` in the analysis configuration files would need to be updated, if they use this implementation
the Run3 trigger sf are in `weights_run3.py`, so in a config, one would need to do this
```python
from pocket_coffea.lib.weights.common.weights_run3 import SF_ele_trigger

weights_classes = [SF_ele_trigger]
```

defaults in the `lepton_scale_factors.yaml` are set for `name` and `file`, `path` needs to be specified by the user

i was thinking about adding a section to the documentation
## Lepton Scale Factors

### Electron

```yaml
trigger_sf:
  year: # e.g. 2022_preEE
    name: JSON Map Name ("Electron-HLF-SF"), has default
    file: Path to JSON file (cvmfs POG/EGM), has default
    path:
      Working Point (e.g. "HLT_SF_Ele30_MVAiso80ID"), needs to be set by the user!
```

what do you think?

felixzinn:
i think thats a good idea 👍🏼

PR #221, opened 2024-08-26
feat: compare option for plotting, allowing shape comparison instead of data/mc

andreypz:
Adding a new feature, enabling the comparison of the MC shapes, instead of doing data/mc. 

In addition, added `poisson_interval()` method for better uncertainties on the ratios. Ratios with `--density` also work now.


andreypz:
Added documentation for the new feature. Should be good to go @valsdav

PR #227, opened 2024-09-03
Correct under/overflow bins

mmarchegiani:
New features:
- **Bugfix:** Correct computation of underflow and overflow bins in MC and systematic uncertainty
- **Bugfix:** Improved defaults for the `Style` class and sanity checks
- **Enhancement:** Generalization of the categorical axes parameters for data and MC with defaults
- **Docs:** Updated documentation

**Correct underflow and overflow bins**
The computation of the MC yield and systematic uncertainty for the underflow and overflow bins is now correctly computed. The underflow and overflow bins are correctly summed to the first and last bin respectively.
The present default is that the user has to set both `underflow` and `overflow` axis parameters to the same value, i.e. the configurations with `underflow=True` and `overflow=False` and viceversa are not supported. However, this can be changed in the future for better flexibility.

**Improved defaults for `Style`**
The defaults for the `Style` object are now taken from the central default parameters of PocketCoffea, from `pocket_coffea/parameters/plotting_style.yaml`. In this way, if the user does not specify any required parameter needed for plotting, the default value is used.

**Generalization of categorical axes**
Two new dictionaries `categorical_axes_data` and `categorical_axes_mc` are introduced as attributes of the `Style` object to parametrize the categorical axes in data and MC histograms and included in the default parameters in `pocket_coffea/parameters/plotting_style.yaml`:
```
plotting_style:
     categorical_axes_data:
          year: years
          cat: categories
      
     categorical_axes_mc:
          year: years
          cat: categories
          variation: variations
```
This feature is needed in order for the user to include additional custom axes in the plotting, for example to include the `era` axis for data. This is done by overwriting the default parameters with the usual `--overwrite-parameters` option of the `make_plots.py` script, redefining the default dictionary such as:
```
plotting_style:
     categorical_axes_data:
          era: eras
```
At present only `era` is supported as additional custom axis for data, while no additional axis is supported for MC. However, in case of future developments the newly supported axes can be specified in the `_available_categorical_axes` method of the `Style` class.

**Updated documentation**
New section for the definition of custom categorical axes in the plotting.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -166,7 +166,7 @@ git clone git@github.com:PocketCoffea/PocketCoffea.git
 
       ```bash
       # Install micromamba
-      curl micro.mamba.pm/install.sh | bash
+      "${SHELL}" <(curl -L micro.mamba.pm/install.sh)
       micromamba create -n pocket-coffea python=3.9 -c conda-forge
       micromamba activate pocket-coffea
       ```
--- a/docs/plots.md
+++ b/docs/plots.md
@@ -159,3 +159,17 @@ plotting_style:
 
 If no alias or default `matplotlib` color corresponds to the string specified by the user,
 an exception is raised.
+
+## Additional custom axes
+
+In order to include an additional custom axis in the plotting, one has to specify the dictionary of categorical axes for data and MC separately.
+For example, to include an additional axis for data to keep track of different data-taking eras as an additional category, one can overwrite the custom `.yaml` file as follows:
+
+```
+plotting_style:
+
+    categorical_axes_data:
+        era: eras
+```
+
+where the key (`era`) in the dictionary corresponds to the name of the axis as saved in the histogram, while the value (`eras`) corresponds to the name assigned to the corresponding attribute of the `Shape` object.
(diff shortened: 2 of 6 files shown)


andreypz:
While I did not test the PR, from a quick look I have one comment. 
[This change](https://github.com/PocketCoffea/PocketCoffea/pull/227/files#diff-40e5adbb0d6ad96f6eb89c8118f8b6321532129b89d9fda801fa1745e746fd89L709-L715) will not propagate stat uncertainty correctly. Which means that with the `--density`option the uncertainty on the ratio will not be correct. That's the reason why I changed the behavior recently.

mmarchegiani:
Hi @andreypz, the only change that I made here is to compute the integral considering also the overflow and underflow bins (which were mistakenly not considered previously) and I propagated the changes to the `if self.density` assuming that the computation below was correct.
Do you have a patch to correct the uncertainty on the ratio for the `density=True` plot?

andreypz:
So, I think you should test this, but the problem is that these two are not equivalent:
```       
     hden = hden * (1./den_integral)
      den = den * (1./den_integral)
```
where `hden` is a Hist object, and `den` is the `values()`. When you scale the hist object the uncertainty is propagate to the variance. When you just scale the values(), the variances are not changed. This leads to wrong uncertainty when taking ratios.

mmarchegiani:
I get what you mean.
Actually at the moment we are saving `num_variances` and `den_variances` to manipulate them independently from the bin values `num` and `bin`. But as you say, at the moment we are not scaling the variances in the case of the `density` plot.
I think this is easily solvable by scaling `num_variances` and `den_variances` by `(1./integral)**2` in the density plot, since the variance scales quadratically with the scaling factor.

andreypz:
UPD: I figured out how to fix this myself. Will make a MR with this myself, soon

PR #229, opened 2024-09-06
FIx the flow for compare option

andreypz:
After #227  was merged, the `--compare` feature was broken. This PR fixes it.
There are also some minor changes to docs.


mmarchegiani:
Thank for the fix in `plot_utils.py`! The PR looks good to me.

PR #232, opened 2024-09-12
Feature/testing

felixzinn:
The purpose of this PR is to introduce style checks and tests into the development workflow.
The pre-commit config and pyproject.toml files have been updated:
- setup pre-commit hooks
- ruff (https://docs.astral.sh/ruff/) for linting and formatting
- mypy for type checking
- update nox configuration

A (draft) CONTRIBUTING.md file has been added.

We should discuss what else we need in pre-commit/nox. It would be great if other people check if the setup works or if I missed something.


valsdav:
Hi @felixzinn, thanks a lot for these needed improvements. 

I'm testing them and then we can add things iteratively when needed.

felixzinn:
i think it would make sense to run formatter/linter for every PR
what do you think?

also what are your thoughts on moving to ruff? or sticking with black/flake8/isort, i have no preference there

PR #233, opened 2024-09-12
Basic methods to deal with GenParticles and GenJets, and more tagger options for Jets

andreypz:
Adding some tools for generator-only analysis. Namely, selecting leptons from GenParticles and selecting GenJets

Plus, a few  other minor changes: `delta_phi` function; new histograms for gen objects, new histograms for PNet and RobustParT tagger scores; minor changes for plotting with --compare option.

### DIFF
--- a/pocket_coffea/lib/gen_objects.py
+++ b/pocket_coffea/lib/gen_objects.py
@@ -26,12 +26,12 @@ def Gen_leptons(events, lepton_flavour, params):
 
 
 
-def Gen_jets(events, GenLepCollName, params):
+def Gen_jets(events, GenLepCollNameForCleaning, params):
 
     cuts = params.object_preselection.GEN_Jet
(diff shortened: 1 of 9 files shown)


andreypz:
@valsdav this is ready for review and merge, please take a look.

pocket_coffea/lib/gen_objects.py
```diff
@@ -0,0 +1,39 @@
+import awkward as ak
+import numpy as np
+
+def Gen_leptons(events, lepton_flavour, params):
```
valsdav, line 4: it may just be my preference, but what about renaming the function `getGenLeptons`?   `GenLeptons` sounds like a class a not a function :)

andreypz: Sure, this change is fine with me. 
However, unless there are other suggestions, I would prefer to change it in my next PR. 
I have modified my code on top of this PR, and don't not want to revert for a small fix like this.

valsdav: sure nothing blocking! Thanks

PR #234, opened 2024-09-20
Introducing dask for RWTH executor

andreypz:
A working version of dask at RWTH linux cluster introduced.

Also, renamed get function for generator leptons and jets.


valsdav:
Hi @andreypz. Sorry I missed that this was ready. Merged!

PR #252, opened 2024-10-14
Preliminary implementation of nominal ele scale smearing

phnattla:
Added a (very preliminary) implementation of nominal ele scale smearing. For working implementation of shifts, the handling of the shape variations needs to be changed, since the nominal values need to be saved.

Work in progress, iterating with @valsdav

### DIFF
--- a/pocket_coffea/lib/leptons.py
+++ b/pocket_coffea/lib/leptons.py
@@ -36,7 +36,7 @@ def get_ele_scaled(ele, jsonFileName, isMC, runNr):
         ele_pt_scale_down = ak.unflatten((1-scale_MC_unc) * ele_pt_flat, counts=ele_counts)
         return {"Up": ele_pt_scale_up, "Down": ele_pt_scale_down}
 
-def get_ele_smeared(mc_ele, jsonFileName, isMC, nominal=True):
+def get_ele_smeared(mc_ele, jsonFileName, isMC, nominal=True, seed=125):
     
     if not isMC:
         return
@@ -46,7 +46,7 @@ def get_ele_smeared(mc_ele, jsonFileName, isMC, nominal=True):
     ele_r9_flat = ak.flatten(mc_ele["r9"])
     ele_pt_flat = ak.flatten(mc_ele["pt"])
     ele_counts = ak.num(mc_ele["pt"])
-    rng = np.random.default_rng(seed=125)
+    rng = np.random.default_rng(seed=seed)
     rho = evaluator_smearing.evaluate(
         "rho",
         ele_eta_flat,
--- a/pocket_coffea/parameters/lepton_scale_factors.yaml
+++ b/pocket_coffea/parameters/lepton_scale_factors.yaml
@@ -95,7 +95,7 @@ lepton_scale_factors:
       "2023_preBPix": "2023PromptC"
       "2023_postBPix": "2023PromptD"
       
-    apply_eleSS: 
+    apply_ele_scale_and_smearing: 
       '2022_preEE': True
       '2022_postEE': True
       '2023_preBPix': True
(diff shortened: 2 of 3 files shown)


pocket_coffea/parameters/lepton_scale_factors.yaml
```diff
@@ -95,6 +95,11 @@ lepton_scale_factors:
       "2023_preBPix": "2023PromptC"
       "2023_postBPix": "2023PromptD"
       
+    apply_eleSS:
```
valsdav, line 98: can we rename it "apply_ele_scale_and_smearing"?

pocket_coffea/workflows/base.py
```diff
@@ -703,6 +704,31 @@ def get_shape_variations(self):
...
+                        self.events["Electron"], etaSC, "etaSC"
+                    )
+                    self.events["Electron"] = ak.with_field(
+                        self.events["Electron"], self.events["Electron"]["pt"], "pt_original"
+                    )
+                    ssfile = self.params.lepton_scale_factors["electron_sf"]["JSONfiles"][self._year]["fileSS"]
+                    # Apply smearing on MC, scaling on Data
+                    if self._isMC:
+                        ele_pt_smeared = get_ele_smeared(self.events["Electron"], ssfile, self._isMC, nominal=True)
+                        self.events["Electron"] = ak.with_field(
```
valsdav, line 719: this can also be expressed with `self.events["Electron", "pt"] =..` but I don't know which one is faster/better

pocket_coffea/lib/leptons.py
```diff
@@ -1,5 +1,74 @@
...
+    
+    if not isMC:
+        return
+    evaluator = correctionlib.CorrectionSet.from_file(jsonFileName)
+    evaluator_smearing = evaluator["Smearing"]
+    ele_eta_flat = ak.flatten(mc_ele["etaSC"])
+    ele_r9_flat = ak.flatten(mc_ele["r9"])
+    ele_pt_flat = ak.flatten(mc_ele["pt"])
+    ele_counts = ak.num(mc_ele["pt"])
+    rng = np.random.default_rng(seed=125)
```
valsdav, line 49: I agree that this needs to be deterministic, but like this the smearning factor is the same at each function invocation. 

Maybe we can create a seed from the dataset/file name? So that it is still reproducibile but different for different files?

phnattla: Do you have a suggestion on how to do this?
Is the function called once per dataset/file, or once per chunk?

valsdav: The function is called once per chunk. so it can be called multiple time per file. 
We can make it dependent only on the filename or on the filename and chunk.

We may use the filename and entrystart to define a seed. Something like
 ```seed = hash(self.events.metadata['fileuuid'])+self.events.metadata['entrystart']```

The result would be reproducible in case the analysis is re-run with the same chunksize. But this is expected with smearing operations. I think it is already nice to be able to reproduce exactly the same numbers if one does not change the file splitting. 

What do you think?

pocket_coffea/parameters/lepton_scale_factors.yaml
```diff
@@ -95,6 +95,11 @@ lepton_scale_factors:
       "2023_preBPix": "2023PromptC"
       "2023_postBPix": "2023PromptD"
       
+    apply_ele_scale_and_smearing:
```
valsdav, line 98: Last request: can you please add also Run2 data taking period identifiers with 'False'?

valsdav:
Thansk a lot! I think we can merge this as soon as you add a default params for Run2 datataking periods ( no SS)

valsdav:
Thanks @phnattla !

Issue #250, opened 2024-10-10
[Plotting] Blind certain histogram bins


felixzinn:
It might be useful to blind certain bins of histograms depending on signal-to-background ratio or some other metric
configurable parameters might be
- metric S/B, S/sqrt(B), ...
- threshold

PR #254, opened 2024-10-15
blind specific histogram bins

felixzinn:
I just opened this PR for the purpose of a discussion (addressing #250 ).
The changes are not yet ready to be merged.

i implemented a simple S/sqrt(B) threshold to blind histogram bins that are above the blind_threshold.

So far there is no good method to know which processes are considered signal or background. i used the `signal_samples` parameter, which however also plots the signal processes as lines.
I think this implementation would benefit from a `Process` class (see #236).


andreypz:
To me the idea to blind bins based on some criteria seems a bit too fine tuned, an unnecessary complication to the code. 
Practically, it is easy to simply mask the bins where you know the signal is high. Such feature is already implemented: https://pocketcoffea.readthedocs.io/en/stable/plots.html#plotting-parameters

Issue #255, opened 2024-10-15
Default Plotting Parameters


felixzinn:
So far there is a [plotting_style.yaml file](https://github.com/PocketCoffea/PocketCoffea/blob/d051d44dd4a2a2e690c282ba04e0c25cefbfee5a/pocket_coffea/parameters/plotting_style.yaml) with defaults.
But in the [Style Class](https://github.com/PocketCoffea/PocketCoffea/blob/d051d44dd4a2a2e690c282ba04e0c25cefbfee5a/pocket_coffea/utils/plot_utils.py#L59-L153) there are some defaults set as well.

it would be better to:
- define all parameters in the yaml file
- load the default yaml file in the Style Class
- overwrite parameters set in user plotting config

felixzinn:
solved in #256

PR #256, opened 2024-10-16
load plotting style defaults from yaml file

felixzinn:
solves #255

the default parameters for plotting are set in the plotting_style.yaml file. In the initialization of the Style class, the yaml file is loaded and the parameters are updated with the user-defined values.


mmarchegiani:
Thanks @felixzinn for this patch.
I think it's safer to load the default plotting parameters from the central .yaml file, and each time a new default is defined, we just updated `plotting_style.yaml`.
The PR looks good to me.

PR #268, opened 2024-11-04
Implemented MET-xy corrections for Run2 UL

guanfacin24:
I have added MET-xy corrections for Run2 UL, based on the current JME recommendations. To correct MET-xy, one needs to add the following lines (or similar) to their workflow:
```
from pocket_coffea.lib.jets import met_xy_correction
met_pt_corr, met_phi_corr = met_xy_correction(self.params, self.events, self._year, self._era)
self.events["MET"] = ak.with_field(self.events["MET"], met_pt_corr, "pt_corr")
self.events["MET"] = ak.with_field(self.events["MET"], met_phi_corr, "phi_corr")
```
I have tested the functionality of the corrections, but did not run any deeper tests (yet, if they are necessary).


valsdav:
Thanks a lot for the feature!

Maybe can you just add a section to document that this feature is available in the `HowTo` documentation page please? I would like to start populating that.

valsdav:
Thanks a lot! Good to go!

PR #276, opened 2024-11-13
Define tagger scores based on used tagger only once when selecting jets, and prevent repeated codes

mondalspandan:
I propose to define the tagger scores as B, CvL, and CvB only once when selecting the jets (similarly BB/CC for fatjets). This will be based on which tagger is specified as a string.

This prevents determining the tagger scores repeatedly in other functions. It is reverse compatible with `jet_selection` itself, but not with the other tagger-dependent functions without minor changes to how you call them (basically just need to remove the tagger name). I have added errors to warn about the deprecation.


mondalspandan:
Hi @valsdav, can you let me know if you agree with this proposal? Thanks!

CC: @andreypz

valsdav:
Hi @mondalspandan sorry for the delay. Having a closer look

valsdav:
How this change is linked to the default btagging parameters?
I think it is a good idea, and then the user can specify the name in the parameters. Of course that is free form and in the processor apply_object_preselection the user can handle the tagger name without constraints.

mondalspandan:
Sorry, what do you mean by default btagging parameters?

valsdav:
I was wondering if we can link this change to a more general config of the btagging WP for different taggers [here](https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/parameters/btagging.yaml), but yeah this PR is not related to this, but more to cache the btag score.

PR #287, opened 2024-12-02
Adding export of columns for shape variations

valsdav:


valsdav:
Related to #339. I noticed I opened a PR about that few months ago but didn't finalize it. 

@leyva-daina if you have time to test it further we can solve the issue :)

PR #288, opened 2024-12-03
Update JEC txt files

gbonomel:
Upload of updated version V2 of JEC txt files taken from [JECDatabase](https://github.com/cms-jet/JECDatabase/tree/master) for Summer22 and Summer22EE MC. The new txt files are saved in `pocket_coffea/parameters/jec` into the respective folders. 

Also the file `pocket_coffea/parameters/jets_calibration.yaml` is modified to use the newly introduced JEC txt files for `2022_preEE` and `2022_postEE`.

PS. From a preliminary check the version V2 of the JEC files shows no difference with respect to the version V1 used until now for Summer22 and Summer22EE MC. This is currently under investigation.


valsdav:
For the records:  the V2 files for DATA are different from V1, but they were already updated in the repo. We are updating the MC files for consistency (but they have the same content)

PR #298, opened 2024-12-10
Parton shower and QCD scale weights

mmarchegiani:
The generator level weights to account for the variation of the QCD scale both in the hard process and in the parton shower (PS) are implemented. In particular, the following weights are implemented:
- PS weight to vary up/down the QCD renormalization scale in the initial-state-radiation (ISR) parton shower
- PS weight to vary up/down the QCD renormalization scale in the final-state-radiation (FSR) parton shower
- Weight to vary up/down the QCD renormalization scale ($\mu_R$) in the hard process
- Weight to vary up/down the QCD factorization scale ($\mu_f$) in the hard process

The weights are taken from the generator-level variables `PSWeight` and `LHEScaleWeight`.

Since the implementation of the PS weights is compatible both with NanoAODv9 and NanoAODv12, it is included in the common weights in PocketCoffea.
Instead, since the implementation of the QCD renormalization and factorization scale weights is specific to NanoAODv9, it is included in the common weights for Run2.

In addition, there is a small patch to the `Configurator` in order to catch the exception when the user is using the weight `bysample` but does not specify either of the `inclusive` or `bycategory` keys.

Reference for PS uncertainties:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/TopSystematics#Parton_shower_uncertainties

Reference for QCD scale uncertainties:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/TopSystematics#Factorization_and_renormalizatio

### DIFF
--- a/pocket_coffea/lib/scale_factors.py
+++ b/pocket_coffea/lib/scale_factors.py
@@ -471,3 +471,29 @@ def sf_pileup_reweight(params, events, year):
     sfdown = puWeightsJSON[puName].evaluate(nPu, 'down')
 
     return sf, sfup, sfdown
+
+def sf_isr(events):
+    '''Up and down variations for the ISR parton shower weights.
+    In order to properly store the weights, a dummy weight of 1 is stored
+    as central value for the ISR correction.
+    Conventions for the PS weights are:
+    [0] is ISR=2 FSR=1; [1] is ISR=1 FSR=2[2] is ISR=0.5 FSR=1; [3] is ISR=1 FSR=0.5;
+    '''
+    isr_up = events.PSWeight[:,2]
+    isr_down = events.PSWeight[:,0]
+    nom = ak.ones_like(isr_up)
+
+    return nom, isr_up, isr_down
+
+def sf_fsr(events):
+    '''Up and down variations for the FSR parton shower weights.
+    In order to properly store the weights, a dummy weight of 1 is stored
+    as central value for the FSR correction.
+    Convention for the PS weights are:
+    [0] is ISR=2 FSR=1; [1] is ISR=1 FSR=2[2] is ISR=0.5 FSR=1; [3] is ISR=1 FSR=0.5;
+    '''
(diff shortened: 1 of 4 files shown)


pocket_coffea/lib/scale_factors.py
```diff
@@ -471,3 +471,29 @@ def sf_pileup_reweight(params, events, year):
     sfdown = puWeightsJSON[puName].evaluate(nPu, 'down')
 
     return sf, sfup, sfdown
+
+def sf_isr(events):
```
valsdav, line 475: better to add parton shower in the name:  in my opinion better  `sf_partonshower_isr`  than `sf_ps_isr`.

pocket_coffea/lib/weights/common/common.py
```diff
@@ -13,6 +13,8 @@
     sf_jet_puId,
     sf_L1prefiring,
     sf_pileup_reweight,
+    sf_isr,
```
valsdav, line 16: same comments for the naming

pocket_coffea/lib/weights/common/weights_run2_UL.py
```diff
@@ -85,9 +85,51 @@ def sf_ele_trigger(params, events, year, variations=["nominal"]):
     return sf_dict
 
 
-########################################################################
-# More complicated WeightWrapper defining dynamic variations depending
-# on the data taking period
+def sf_renscale(events):
```
valsdav, line 88: Just a comment on the naming: `sf_renorm_scale`?  Just to make it a bit clearer

pocket_coffea/lib/scale_factors.py
mmarchegiani: Done

pocket_coffea/lib/weights/common/common.py
mmarchegiani: Done

valsdav:
Thanks Matteo!

PR #303, opened 2025-01-16
Add sorting by b tagging score AND fix naming convention of 2023 eras in jet veto map config

mondalspandan:


valsdav:
HI! May I ask you to include in this PR the update of the jet veto map that came out today?

https://cms-talk.web.cern.ch/t/updated-jetvetomap-for-summer23bpixprompt23/97635/1

mondalspandan:
Hi, I checked with JERC and this is what I gather:

In the older version, we were supposed to do:
`corr.evaluate("jetvetomap", etaFlat, phiFlat) - corr.evaluate("jetvetomap_bpix", etaFlat, phiFlat)`
in 2023_postBPix but [it seems the second term was not implemented in the first place](https://github.com/PocketCoffea/PocketCoffea/blob/2d6f00b0badf4513ed08f3b0f2a0c5f2e7b315f7/pocket_coffea/lib/cut_functions.py#L95).

In the newer version of the vetomaps, this subtraction is not necessary and 
`corr.evaluate("jetvetomap", etaFlat, phiFlat)` is enough.

The [`jsonpog-integration` files](https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/blob/master/POG/JME/2023_Summer23BPix/jetvetomaps.json.gz?ref_type=heads) have already been updated by JME. So the current code in PoCo already retrieves the updated vetomap and correctly applies the veto based only on `jetvetomap`.

valsdav:
Ok perfect, thanks for double checking.

PR #307, opened 2025-01-29
Added MET rescaling after JEC

### LINKED ISSUE #186: Propagate JEC correction to MET
In Run3,  JEC is reapplied on top of NanoAOD in the framework: we need to propagate the correction to the MET variables.

valsdav:
MET needs to be rescaled when updating the JEC on the fly. 

The correction is added in  `get_shape_variations` loop, both for nominal JEC and for varied JES.

N.B.: this is working both for the nominal JEC application and variations. 

Fixes #186


pocket_coffea/lib/jets.py
```diff
@@ -46,13 +46,33 @@ def jet_correction(params, events, jets, factory, jet_type, chunk_metadata, cach
...
+def met_correction_after_jec(events, METcoll, jets_pre_jec, jets_post_jec):
+    '''This function rescale the MET vector by minus delta of the jets after JEC correction
+    and before the jEC correction.
+    This can be used also to rescale the MET when updating on the fly the JEC calibration. '''
+    orig_tot_px = ak.sum(jets_pre_jec.px, axis=1)
+    orig_tot_py = ak.sum(jets_pre_jec.py, axis=1)
+    new_tot_px = ak.sum(jets_post_jec.px, axis=1)
+    new_tot_py = ak.sum(jets_post_jec.py, axis=1)
+    new_met_px = events[METcoll].px - (new_tot_px - orig_tot_px)
+    new_met_py = events[METcoll].py - (new_tot_py - orig_tot_py)
```
mmarchegiani, line 58: `new_met_px` and `new_met_py` are defined here but are not used later.
This is a duplicate of the two lines below right?

valsdav: Ah yes I think this is a duplicate

PR #314, opened 2025-02-12
Minor changes to executor parameters, rucio script and requirements

andreypz:
These are some minor changes for better running at RWTH:
* Fix typos in default parameters, and set memory to 4Gb by default
* In `rucio.py` script increased the validity of the `sitemap` file to 60 minutes. Twenty minutes was no longer enough, because the script actually takes longer than 20 minutes to finish! (we have a lot of samples to build).
* Added `dask` to requirements.txt. It is actually needed if installing with mamba/conda, otherwise dask is never installed.


requirements.txt
```diff
@@ -19,3 +19,5 @@ law
 click>=8.0.4
 correctionlib>=2.6.1
 setuptools<71
+dask
```
valsdav, line 22: Hi @andreypz on this: the problem is that this would setup dask in the wrong way in our pipeline to build the apptainer images. We can add them as an opional requirements in the pyproject.yaml. What do you think about that?

andreypz: Ok, sure. I removed dask from the list.
I'm don't know how the pyptoject file is used actually.
But my point is that for mamba/conda users who do not use the apptainer image it needs to be installed. Maybe just writing this in the docs would be enough.

valsdav: You can a line in `pyproject.toml` 

```
[project.optional-dependencies]
dev = ["snakeviz", "memray", "flake8", "black", "pytest>=6", "pytest-cov>=3", "pre-commit", "nox"]
docs = ["sphinx>=4.0", "myst-parser>=0.13", "sphinx-book-theme>=0.1.0", "sphinx-copybutton", "sphinx-markdown-tables", "nox"]
test = ["pytest>=6", "pytest-cov>=3"]
dask = ["dask>xxx"]
```

so that `pip install .[dask]` will install also those directly :+1:

andreypz: added the dask option to pyprojct

andreypz:
Could you please merge. I have more dev on top of this

valsdav:
Thanks @andreypz !

PR #323, opened 2025-03-09
Modify the make-plot script & Add UMD_T3 executor

De-Cristo:
Dear developer,

This commit contains some updates that we have discussed in the mattermost channel, which includes:

1. Modify the `pocket_coffea/scripts/plot/make_plots.py` script, so that it can accept multiple inputs files by doing `-i A.coffea -i B.coffea...`
- the thing I haven't found proper solution for now is to accept wildcards. It might be easy but I found it quite difficult to do it with `click.option` .

2. Add the CMS_UMD_T3 cluster executors [Commit: 7a75a227b4737d2a26cc3f1878f86647286b4eec] for both condor and dask, we name it `condor@rubin` and `dask@rubin`. The modification includes: `docs/running.md`, `pocket_coffea/executors/executors_rubin.py`, `pocket_coffea/parameters/executor_options_defaults.yaml`, `pocket_coffea/scripts/runner.py`. Please check if the commit misses anything necessary and also it is not disturbing the existing workflow.

Best,
Licheng

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -120,6 +120,7 @@ respectively).
 |[INFN Analysis Facility](https://infn-cms-analysisfacility.readthedocs.io/)| dask | dask@infn-af |
 |Brown brux20 cluster | dask | dask@brux |
 |Brown CCV Oscar | dask | dask@oscar |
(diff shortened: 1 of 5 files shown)


valsdav:
Thanks @De-Cristo

PR #330, opened 2025-04-01
Generalize object calibrations and variations

valsdav:
This PR introduces `Calibrator` object and the `CalibratorsManager` to handle in a more general way the calibration of collections and their corresponding "shape" uncertainties.

Similarly to the Weights classes configuration, the user can now customize the calibration of collection and provide custom Calibrator objects to the configurator. The framework will use the calibrator objects in a sequence to adjust the event collection and customize the available shape variations for each chunk of events. In this way the variations can be customized by datataking period and other medatata. 

The base workflow is simplified and all the hardcoding  for collection calibration is removed. The CalibratorsManager handles the application of the Calibrator objects in sequence over the events.   Doing so the user can also create custom calibrations by manipulating directly the calibrator sequence.  A set of default calibrators is made available in the framework for JEC, electron scale&smearing, MET correction.


pocket_coffea/lib/calibrators/common/pnet_regression.py
```diff
@@ -0,0 +1,101 @@
...
+        required_fields = ['rawFactor', 'PNetRegPtRawCorr', 'PNetRegPtRawCorrNeutrino', 'btagPNetB', 'btagPNetCvL']
+        missing_fields = [field for field in required_fields if field not in j_flat.fields]
+        
+        if missing_fields:
+            # If required fields are missing, return original jets without regression
+            out["Jet"] = jets
+            return out
+        
+        # Obtain the regressed PT and Mass
+        reg_j_pt = j_flat['pt'] * (1 - j_flat['rawFactor']) * j_flat['PNetRegPtRawCorr'] * j_flat['PNetRegPtRawCorrNeutrino']
```
valsdav, line 82: need to be careful pt > 15 and eta < 4.7 and check RawCorr != 0

PR #332, opened 2025-05-16
Update/plotting logscale

felixzinn:
in #331 the option to plot both x *and* y axis in log-scale has been added
this PR also updates the corresponding law task parameters


pocket_coffea/law_tasks/tasks/plotting.py
```diff
@@ -61,6 +93,12 @@ def setup_plot_manager(self):
             for key in vars_to_pop:
                 output_coffea["variables"].pop(key, None)
 
+        if not output_coffea["variables"]:
```
felixzinn, line 96: if the regex does not match any variables in the coffea output file, an error will be raised
(was not really worth a separate PR so included it here)

valsdav:
Thanks Felix!

PR #338, opened 2025-06-02
Update recipes.md to add onnx executor how-to

Emma-223:
Adds instructions in the How-To section describing how to use onnxruntime in a custom executor.


valsdav:
Thanks a lot @Emma-223 . 

There are many things we can improve on the ML inference workflow but already having the basic example documented is a very good step forward.

valsdav:
https://pocketcoffea.readthedocs.io/en/latest/recipes.html#create-a-custom-executor-to-use-onnxruntime

Issue #344, opened 2025-06-23
Update Coffea to 2025


felixzinn:
in the new coffea v2025.5.0 introduces (among other stuff)
- virtual arrays in awkward 2
- 0.7 style processor/executor
see release notes (https://github.com/scikit-hep/coffea/releases/tag/v2025.5.0.rc0)

we should keep an eye on that

PR #351, opened 2025-07-24
coffea: update coffea version to 2025.07

felixzinn:
Hi,

i was testing out the new coffea 2025.07 version https://github.com/scikit-hep/coffea/releases/tag/v2025.7.0
i did some small tests with very little syntax changes for the tests to complete
I open this PR to work on it and discuss

related #344


valsdav:
Hi Felix, this is great thanks for starting this! 

May I ask you to push the branch also on the gitlab repository so that we can run the CDCI and test this?

pocket_coffea/lib/calibrators/calibrators_manager.py
```diff
@@ -133,7 +136,7 @@ def calibrate(self, events, variation, debug=False):
                     if col not in self.original_coll:
                         try:
                             self.original_coll[col] = events[collection, field]
-                        except ValueError:
+                        except ak.errors.FieldNotFoundError:
```
felixzinn, line 139: please confirm, that this is the correct logic @valsdav 

as I understand the field will be filled if it does not exist
awkward raises `FieldNotFoundError` in that case

PR #352, opened 2025-07-29
Filter invalid files in a dataset

andreypz:
It could happen sometimes that a Valid dataset contains a few invalid files.
Example in `/Zto2Nu-2Jets_PTNuNu-200to400_2J_TuneCP5_13p6TeV_amcatnloFXFX-pythia8/Run3Summer22EENanoAODv12-130X_mcRun3_2022_realistic_postEE_v6-v1/NANOAODSIM` dataset, where one file is invalid.

Rucio does not know about invalid file, but DAS knows.

This PR allows to filter such files.

### DIFF
--- a/pocket_coffea/utils/dataset.py
+++ b/pocket_coffea/utils/dataset.py
@@ -173,6 +173,7 @@ class Sample:
                 verify=False,
             )
             filesjson = r.json()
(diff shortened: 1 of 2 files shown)


valsdav:
Thanks for the nice fix!

PR #355, opened 2025-08-04
Jet selection fix for NanoAODv12

gbonomel:
A new definition for the jet object has been implemented, especially only for NanoAOD v12 after the known [issue](https://gitlab.cern.ch/cms-jetmet/coordination/coordination/-/issues/117) in the jetId. 
This code contains a custom function to select JetGood following the recommendations from JET MET.

### DIFF
--- a/pocket_coffea/lib/jets.py
+++ b/pocket_coffea/lib/jets.py
@@ -374,17 +374,15 @@ def jet_selection_nanoaodv12(events, jet_type, params, year, leptons_collection=
         (jets.pt > cuts["pt"])
         & (np.abs(jets.eta) < cuts["eta"])
     )
-    passJetIdTight = ak.zeros_like(jets.pt, dtype=bool)
-    passJetIdTightLepVeto = ak.zeros_like(jets.pt, dtype=bool)
-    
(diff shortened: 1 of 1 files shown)


pocket_coffea/lib/jets.py
```diff
@@ -363,6 +363,96 @@ def jet_selection(events, jet_type, params, year, leptons_collection="", jet_tag
...
+    # Bug fix for NanoAOD v12 https://gitlab.cern.ch/cms-jetmet/coordination/coordination/-/issues/117
+    
+    mask_presel = (
+        (jets.pt > cuts["pt"])
+        & (np.abs(jets.eta) < cuts["eta"])
+    )
+    passJetIdTight = ak.zeros_like(jets.pt, dtype=bool)
+    passJetIdTightLepVeto = ak.zeros_like(jets.pt, dtype=bool)
+    
+    passJetIdTight = ak.where((np.abs(jets.eta) <= 2.7), (jets.jetId & (1 << 1))>0, passJetIdTight)
```
valsdav, line 380: Can we also include the cases where the requested jetid in the parameters is looser ? in order to simulate the same behaviour of the `jet_selection` standard function

PR #359, opened 2025-08-07
Update paramers for DESY executor

andreypz:
Here are just some updates to run at DESY (with Parsl), mostly updating for more common parameters (memory, conda-env etc).

Plus some updates to documentation

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -113,7 +113,7 @@ respectively).
 |lxplus| dask               | dask@lxplus    |
 |swan| dask                 | dask@swan    |
 |T3_CH_PSI| dask            | dask@T3_CH_PSI |
-|DESY NAF | dask            | dask@DESY_NAF |
+|DESY NAF | dask,parsl      | dask@DESY_NAF, parsl@DESY |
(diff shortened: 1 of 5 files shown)


andreypz:
Sorry that this change affected you negatively. But I am not sure where would I announce this beforehand. Perhaps you can sign up for PR notifications for PocketCoffea?

Regarding the changes, I think the `parsl@DESY` is just simpler, and consistent with other sites.

As for the certificate change, sorry I did not realize there was a real use case for not having a certificate. Please add this check back (or I could do it in my next PR, if you want). On the other hand, does it really crash because of this change? It seems that the x509 path should  be just set to empty, no?

andreypz:
UPD
Ok, I see why it would probably crush without  `ignore-grid-certificate` check. In [this line](https://github.com/PocketCoffea/PocketCoffea/blob/cccc8ba85c4156426c12132e5f0897092d28bd19/pocket_coffea/executors/executors_base.py#L23) of the base executor, the function just returns, never setting the `self.x509_path` variable. 
Perhaps the minimalist change would be to set `self.x509_path = ""` at this point? Then we would not need any other checks in the site-specific executors.

PR #362, opened 2025-08-09
Add split-by-category feature for handling large outputs with histograms from a large number of categories

mondalspandan:
The main feature addition is splitting outputs into different categories so that large .coffea files need not be handled in memory during plotting. This enables analyses that have many categories which makes the `merge-output` script fail.

New features:

**`condor@lxplus` executor:**

- ~~Unfortunately some overlaps with #358~~ Resolved
- `--split-by-category` option which splits `output_all.coffea` into several chunks, each containing disjoint set of 8 categories. It calls `split-output`, which now has a new feature `--by category`, which in turn calls another new function, `utils.filter_output.filter_output_by_category`.

**Thoughts:** Coding-wise, this would be easier and more natural to handle by directly calling `utils.filter_output.filter_output_by_category` on `output` within `runner`. But this felt more like an executor's job (specifically, only for manual executors) so I decided to make `--split-by-category` an executor option instead of cluttering `runner`.

**`merge-output` script:**

- **Feature 1:** If memory usage exceeds 50% of available memory, an OOM kill is impending. Avoid this by dumping to disk the accumulated result so far. Then proceed with the remaining files. This is necessary when accumulating a large number of `categories × variables × systematics` even while using a small `N_reduction` value, since the accumulated `result` keeps growing in memory when more categories get added by a new file.
- **Feature 2:** If `--split-by-category` was saved as true in the `jobs_config`, it merges per category-group in a for loop instead of merging all files once. Thus it produces as many outputs as category-groups, instead of just one.
- Overall improved memory handling by explicitly deleting intermediate python objects from memory. 

**`make-plots` script:**

- Handle `--split-by-category` (explicit flag added) by calling itself *once per* category-group.

**`plot_utils` script:**

- Better handling of `--verbose`.
- Delete histograms from memory once `Shape` is created.

**Documentation:**

- Added section "Split large outputs into categories" in `recipes`.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -165,3 +165,16 @@ logging:
   distributed.nanny: warning
   distributed.scheduler: warning
 ```
+
+## Split large outputs into categories
+
+If your configuration contains a large number of categories, variables, and systematics, the number of histograms can grow very large. Although on disk, the size of the *merged* `output_all.coffea` is usually less than O(10 GB), the full accumulation process can consume around O(100 GB) of RAM. This seems unavoidable as coffea accumulation necessarily happens on memory. This can be addressed in one of two ways:
+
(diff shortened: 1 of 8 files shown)


valsdav:
Thanks a lot for all the nice tools for very large analyses!

PR #363, opened 2025-08-10
check-jobs: Replace files with xrootd errors; implement blacklist for sites with repeated failures

mondalspandan:
Two main features:

1. For failed files, check the .err to see if there was an XRootD error. In that case, find an alternative XRootD site for the failed file and replace it. Keep a log of XRootD failed paths on disk and avoid them if a second site fails for the same file.
2. If a site records a large number of failed files (10 by default), replace all files from that site within the current config of a failed job and don't use it in future file replacements. Keep a log of blacklisted sites on disk.

Others:
1. Also look into idle jobs when checking aborted job list. Re-resubmission of newly idle jobs is prevented by maintaining a log of resubmitted jobs on disk. 
2. Other minor logging changes. 

Just one caveat:
I could not fully test the blacklist enforcement because all CMS sites are working flawlessly (gasp!) *after* I finished coding that part. Nevertheless, the XRootD part should work out of the box.

### DIFF
--- a/pocket_coffea/scripts/check_jobs.py
+++ b/pocket_coffea/scripts/check_jobs.py
@@ -433,8 +433,6 @@ def check_jobs(jobs_folder, details, resubmit, max_resubmit, blacklist_threshold
                             with open(maxtimefile,'a') as f:
                                 f.write(job_name+"\n")
 
-                            os.system(f"mv {failedlog} {jobs_folder}/logs/processedlogs")
-
                             # Modify the sub file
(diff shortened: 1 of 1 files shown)


mondalspandan:
Added another feature:

Check logs of "running"/"idle" jobs which were resubmitted, for signs of job abortion. If job failed due to periodic remove, bump to longer queue and resubmit.

This utilizes `condor_history`, so relies on the scheduler being active. If it does not get a response, it tries again in the next pass.

PR #364, opened 2025-08-12
Store correctionlib files locally, versioning with tags

valsdav:
At the moment many correctionlib json files are read directly from CMVFS. This may cause sudden changes if the files are updated centrally. 

In Run3 it is expected that corrections will be updated more often. CAT is developing a versioning schema for files on cvmfs, but in the meanwhile we need a temporary solution to avoid moving defaults in PocketCoffea. 

This PR introduces a new mechanism to store a copy of the correctionlib files from cvmfs in the package:
- versioning with tags
- avoid unnecessary files copy by using hard-links for unchanged files
- storing full file metadata with checksum and timestamps
- transparent configuration in the parameters using a new `${cvmfs:/....}` yaml resolver (thanks to OmegaConf)

Tools for downloading the files and reading the default parameters are included.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ recursive-include pocket_coffea/ *.json
 recursive-include pocket_coffea/ *.txt
 recursive-include pocket_coffea/ *.yaml
 recursive-include pocket_coffea/ *.txt.gz
+recursive-include pocket_coffea/ *.json.gz
 prune notebooks
 prune config
 prune docs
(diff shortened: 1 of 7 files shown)


valsdav:
Any comment on this ? 

(@andreypz @felixzinn @phnattla @matteomalucchi @mondalspandan)

felixzinn:
I like the possibility to download files from cvmfs. I've been doing that for a while now, since our workers do not have cvmfs mounted. so 👍🏼 on that aspect

I quickly skimmed through the code and something that is not clear to me is:
the correctionlib files will become part of the PocketCoffea repo right? However there exists already the POG repo on gitlab and I wonder whether there is an easier way and just refer to that repo. Everyone working with PocketCoffea probably also has access to that repository.

PR #367, opened 2025-08-13
Fixed typo for 2022EE smearing lepton sf

thierryharte:
Very small fix of a typo in the `lepton_scale_factors.yaml`.
"Smear" would not be found by PocketCoffea and lead to a crash, if `ElectronsScaleCalibrator`, so the smearing parameter needs to be named "Smearing".


valsdav:
Hi! Thanks for catching this. It is like this only for that data taking period? 

I know that the EGM files are still not all equal for all the year of Run 3 (we are working on it :)  ), there will be a full coherent release of the SF setup later.

PR #368, opened 2025-08-15
Saving variations in subfolders of category

thierryharte:
Enables saving of different variations.

**This implementation does change the structure of the outputs:**
Instead of the columns being saved in a structure:
`Sample->Dataset->Category->Column`
The new structure is:
`Sample->Dataset->Category->Variation->Column`
This is also true if no variation is activated. In which case the "variation" is called `nominal`

A bit more technical:
To achieve, that the `.coffea` file would contain all variations, the `ColumnsManager` does not have its output reset on every `fill_columns_accumulator` anymore but expands on the same output while it is instantiated. This might lead to some issues that I did not see, but so far no unwanted behaviour occurred in my tests.
For the option to save to `parquet`, simply an additional `subdir` is added. This should not pose any issues.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -9,7 +9,7 @@ def events():
     filename = "root://eoscms.cern.ch//eos/cms/store/mc/RunIISummer20UL18NanoAODv9/TTTo2L2Nu_TuneCP5_13TeV-powheg-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1-v1/280000/01881676-C30A-2142-B3B2-7A8449DAF8EF.root"
 
     print(filename)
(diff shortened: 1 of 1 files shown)


pocket_coffea/workflows/base.py
```diff
@@ -497,8 +497,8 @@ def define_column_accumulators_extra(self):
         '''
 
     def fill_column_accumulators(self, variation):
-        if variation != "nominal":
-            return
+        # if variation != "nominal":
```
valsdav, line 500: I think you should remove directly this commented part, as now it would be active

thierryharte: Nice catch, I forgot this comment. 
Removed it now.

PR #379, opened 2025-08-27
Use absolute paths +  argument in split-output

mmarchegiani:
This PR implements a small patch to the new `ExecutorFactoryCondorCERN`.
I had to modify the bash script in `splitcommands`.

Improvements:

* When splitting output by category, the output filename is now explicitly set to `output.coffea` using the `-o` option so that the split outputs will be like `output_CAT1.coffea`, `output_CAT2.coffea` (remove `_all` subscript)
* The split-output command uses the absolute output path for consistency.
* The script now uses the absolute output path (`abs_output_path`) for all output file operations, both when splitting by category and when copying unsplit output files.

These changes are needed following commit https://github.com/PocketCoffea/PocketCoffea/commit/e934caf3ce804d18c58bfba9507eb22b040d380e.

@mondalspandan, could you have a look if the script will work as intended? Thanks!

### DIFF
--- a/pocket_coffea/executors/executors_lxplus.py
+++ b/pocket_coffea/executors/executors_lxplus.py
@@ -186,12 +186,11 @@ class ExecutorFactoryCondorCERN(ExecutorFactoryManualABC):
             runnercmd = "pocket-coffea run"
 
         # Specify output filename to split-output script ->
-        # This will save files such as output_CAT1.coffea, output_CAT2.coffea ...
+        # This will save files such as output_CAT1.coffea, output_CAT2.coffea (remove "_all" from split outputs)...
         if self.run_options["split-by-category"]:
-            output_filename = "output.coffea"
             splitcommands = f'''
                 cd {abs_output_path}
-                split-output output_all.coffea -b category -o {output_filename}
+                split-output output_all.coffea -b category -o output.coffea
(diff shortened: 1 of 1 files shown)


mmarchegiani:
And another question on this bash script:
Do you think it's safe to do `rm output_all.coffea` after the `split-output` script?
I understand the logic that the output will be stored in separate files and one can remove the inclusive file `output_all.coffea` (of O(10 GB)) after splitting, but do we trust 100% the `split-output` script?
I'm thinking if there might be a corner case in which the `split-output` corrupts the output and then the inclusive output is removed with consequent loss of data.
Since `output_all.coffea` can be the result of hours of CPU time, I usually refrain from removing the `output_all.coffea` file right away.
But I guess this was tested and there are no issues with the `split-output` script so far :smiley:

mondalspandan:
The alternative is to copy both the `_all` file and the split files, but that's really 2x more disk space and I/O consumption. Perhaps we can have yet another flag in the future, like `--keep-all` for this purpose in case someone wants to be doubly sure?

One would use `--split-by-category` only when the set of `output_all`s is expected to be too large to accumulate in memory (> ~35 GB in memory after loading all, on lxplus). In such cases, `output_all`s are pretty much useless, so it may not make sense to copy them back.

Issue #383, opened 2025-09-09
There is no protection against copy failures after a condor@lxplus job ends successfully


mondalspandan:
https://github.com/PocketCoffea/PocketCoffea/blob/46a65211a4715ea907ec61847be4f4d338a0bac5/pocket_coffea/executors/executors_lxplus.py#L216

The copy step sometimes fails with an I/O error:

```
cp: failed to close '/eos/user/s/spmondal/.../output_job_1808.coffea': Remote I/O error
```

I can't think of a graceful way to catch this and mark the job as failed. We can brute force it per `cp` line, but that makes the code ugly (`if`s inside if).

Even better, have each cp inside a `while <stdout is 0>`.

valsdav:
We can catch if the cp command succed with : 

```
if [ $? -eq 0 ]; then echo "Success"; else echo "Failure"; fi
```

For copies to EOS xrootd should be used, it's much more stable. 

`xrdcp root://eosuser.cern.ch//eos/user/....`

mondalspandan:
Right, it just becomes too cluttered that way:
```
if [ $? -eq 0 ]; then
    echo 'Job successful'
    cp ....
    if [ $? -eq 0 ]; then
        rm $JOBDIR/job_$1.running
        touch $JOBDIR/job_$1.done
    else
        echo 'Copy failed'
        rm $JOBDIR/job_$1.running
        touch $JOBDIR/job_$1.failed
else
    echo 'Job failed'
    rm $JOBDIR/job_$1.running
    touch $JOBDIR/job_$1.failed
```
I can make this change, just that it's not too graceful :D 

Thanks for the `root://eosuser.cern.ch/` tip, I will switch to this as the output directory in the future!

valsdav:
well this is a auto-generated script, and if we don't catch that the copy was unsuccessfull it is a bit problem imho.

andreypz:
For xrd copy I would recommend using `xrdcp -f` which should overwrite the file if it exists.

mondalspandan:
Addressed in #385

PR #386, opened 2025-09-13
Update check-jobs, merge-outputs scripts; new merge_columns script

mondalspandan:
**Changelog:*

`check_jobs.py`:
- `--queue-shift` argument to specify how many queues to jump while resubmitting time-lapsed jobs. E.g. `--queue-shift 2` would move job from `microcentury` to `workday`. 
- Mark resubmission as success only if the stdout from `condor_submit` makes sense.
- Minor changes in reading log files. 

`merge_outputs.py`
- `--skip-check` argument to let the merging proceed with incomplete set of outputs.
- `--mark_failed` argument to mark all jobs with missing output as failed in the condor job directory. (E.g. job_x.running -> job_x.failed)
- Detect `jobs_config.yaml` automatically if the directory is passed as `-jc`.
- Additional protection against corrupted/zombie output .coffea files, e.g. when the `cp` fails (refer #383), using `os.path.getsize(job['output_file']) < 10`
- Delete `batch` from memory after being loaded.

`merge_columns.py`
- New script to merge column outputs into one parquet file per dataset, per category, by looking recursively within directories.

### DIFF
--- a/pocket_coffea/scripts/check_jobs.py
+++ b/pocket_coffea/scripts/check_jobs.py
@@ -132,14 +132,14 @@ def update_blacklist(xrootdfaillist,blacklist_threshold):
             blacklist_sites.append(site)
     return blacklist_sites
 
-def bump_jobqueue(sub_file):
+def bump_jobqueue(sub_file, shift=1):
     with open(sub_file) as f:
         lines = f.readlines()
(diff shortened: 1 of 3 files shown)


valsdav:
Sorry @mondalspandan , lost track of this.  If this still updated?

PR #416, opened 2025-10-23
Approach to generalise Datacard object

thierryharte:
There will be additional features coming.
Generally they are intended to make the use of these classes possible to our analysis while not breaking any existing behaviour.

- Added a new optional parameter to the `Datacard` class called `process_suffix`. This should work similar to the `bin_suffix` and default to adding the year. This allows process names without the year inside.
- Slightly changed handling of variation names. Now, `systematic.name` is what the variation is called in PocketCoffea and `systematic.datacard_name` is how the variation is referenced in the datacard. Previously in both cases, `systematic.datacard_name` was used which forced usage of the PocketCoffea name in datacard.


pocket_coffea/utils/stat/systematics.py
```diff
@@ -97,9 +97,13 @@ def list_type(self, syst_type: str) -> list[str]:
         return [key for key in self if self[key].typ == syst_type]
 
     def get_systematics_by_type(self, syst_type: str) -> dict[SystematicUncertainty]:
-        """Dict of Systematics of a specific type."""
+        """Dict of Systematics of a specific type by datacard_name."""
         return {name: syst for name, syst in self.items() if syst.typ == syst_type}
 
+    def get_systematics_by_name(self, syst_type: str) -> dict[SystematicUncertainty]:
```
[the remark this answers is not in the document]
thierryharte: Thanks for the feedback:),
Yes, I see that point. The naming is indeed not optimal.
So the context for this new function is, that we need the `name` and not the `datacard_name` in this place:
https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/utils/stat/combine.py#L433
Because here we look for the histogram in the `coffea` file and this is usually a different name than in the datacard.

Thinking about it, defining a new function for that might have been a bit overkill. I see, that we could still use `get_systematics_by_type()` and then call `syst.name`.

I'll try that.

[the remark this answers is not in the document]
felixzinn: ok, I understand
you can also simply iterate over the `Systematics` dictionary directly if that makes it easier

pocket_coffea/utils/stat/combine.py
```diff
@@ -80,9 +81,12 @@ def __init__(
         self.category = category
         self.bin_prefix = bin_prefix
         self.bin_suffix = bin_suffix
+        self.process_suffix = process_suffix
         self.verbose = verbose
         if self.bin_suffix is None:
             self.bin_suffix = "_".join(self.years)
+        if self.process_suffix is None:
+            self.process_suffix = "_".join(self.years)
```
mmarchegiani: Isn't it better to have:
`self.process_suffix = '_' + "_".join(self.years)` ?

otherwise you might have something like`QCD2022_2023`. I think it's better to have an additional `_` between process and suffix.

pocket_coffea/utils/stat/combine.py
```diff
@@ -104,10 +108,10 @@ def __init__(
         for process in self.mc_processes.values():
             for year in process.years:
                 if process.is_signal:
-                    self.process_id[f"{process.name}_{year}"] = id_signal
+                    self.process_id[f"{process.name}{self.process_suffix}"] = id_signal
```
mmarchegiani: Similarly here (and in all occurrences below): should the key be `f"{process.name}{self.process_suffix}"` or `f"{process.name}_{self.process_suffix}"`? Maybe I would prefer the second so the user doesn't need to know that it should append a `_` to the `process_suffix` string.

pocket_coffea/utils/stat/systematics.py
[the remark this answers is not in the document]
thierryharte: This should now be fixed. In the one place we need the actual name, we are now accessing it through the systematics directly.

tests/test_full_configs/test_datacard_creation.py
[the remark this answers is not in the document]
thierryharte: Hi sorry that it took me so long to follow up on this. I will now finally try to finish this PR.
I did now implement the tests how you recommended.
Like you say, there are other tests to ensure, that the coffea ouput file stays the same, then we can just care about the datacard creation alone.

thierryharte:
The tests for the datacards are now finally implemented.
So this PR from my side is ready for reviewing again.
Sorry that it took me so long to come around to this.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 typing-extensions>=3.7;python_version<'3.8'
-coffea==0.7.29
+coffea>=0.7.29,<0.7.31
```
felixzinn, line 2: did you run all test after this? because this will install 0.7.30. Although I dont think there were major changes that break anything

this is not strictly necessary for this PR though right?
also why exclude v0.7.31 here?
https://github.com/scikit-hep/coffea/releases/tag/v0.7.31

tests/test_full_configs/test_datacard_creation.py
```diff
@@ -0,0 +1,101 @@
...
+#     monkeypatch.chdir(base_path / "test_datacard_creation_files")
+
+
+@pytest.fixture(scope="module")
+def run_pc_config_run3(base_path: Path, tmp_path_factory):
+    os.chdir(base_path / "test_datacard_creation_files")
+    if os.path.exists("jets_calibrator_JES_JER_Syst.pkl.gz"):
+        os.remove("jets_calibrator_JES_JER_Syst.pkl.gz")
+    outputdir = tmp_path_factory.mktemp("test_datacard_creation_files")
+    # config = load_config("config_allvars_Run3.py", save_config=True, outputdir=outputdir)
```
felixzinn, line 33: as I understand there is a fixed .coffea file now for the tests, so no processing needed and this comment can be removed?

tests/test_full_configs/test_datacard_creation_files/coffea_file/output_all.coffea
felixzinn, line 1: did you create this coffea file with the config and parameters inside the test_datacard_creation_files directory? e.g. using this config `tests/test_full_configs/test_datacard_creation_files/config_allvars_Run3.py`

pocket_coffea/lib/jets.py
felixzinn, line 1: is this related to this PR, why the change?

its probably a nice idea to provide it in pocketcoffea though

felixzinn (commented):
thanks for the work!
just had a quick look at the PR and left some comments
will try the new interface and tests when I have time

also I think it is a good idea to format the _new_ files with black, so that at least newly introduced files in PocketCoffea are formatted. for already existing files its better to not format them in this PR, so its easy to tell which changes actually come from the new implementation

PR #421, opened 2025-11-03
NanoAOD version parameter 

valsdav:
The NanoAOD version is necessary for some correction, like the jetID. 

The default version for each data taking period is now part of the default parameters (yaml). 

Single datasets can be configured with a specific `nano_version` in the dataset metadata. 

The jetID implementation has been modified to use this info from the parameters. 
The `jet_selection_nanoaodv12` function has been removed. A `compute_jetId` function adds now on the fly the jetID from correction lib (nanoAODv15) or computes it NanoAOD v12/v13

### DIFF
--- a/pocket_coffea/lib/jets.py
+++ b/pocket_coffea/lib/jets.py
@@ -4,9 +4,8 @@ import awkward as ak
 import numpy as np
 import correctionlib
 from coffea.jetmet_tools import  CorrectedMETFactory
-from ..lib.deltaR_matching import get_matching_pairs_indices, object_matching
 from correctionlib.schemav2 import Correction, CorrectionSet
-from ..utils.utils import get_nano_version
+from ..utils.utils import get_nano_version, replace_at_indices
 
 
 def add_jec_variables(jets, event_rho, isMC=True):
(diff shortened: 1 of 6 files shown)


pocket_coffea/parameters/nano_version.yaml
```diff
@@ -0,0 +1,10 @@
+default_nano_version:
+  2016_PreVFP: 9
+  2016_PostVFP: 9
+  "2017": 9
+  "2018": 9
+  '2022_preEE': 12
+  '2022_postEE': 12
+  '2023_preBPix': 12
+  '2023_postBPix': 12
+  '2024': 15
```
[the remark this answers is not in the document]
valsdav: I agree, that can be added, but this is the default_nano_version, in case a dataset does not have the metadata defined. It is a way of defining the default globally. 

Extract the version from the string is doable I think. We can put it in the base.py `load_metadata` function

pocket_coffea/lib/jets.py
Jennyggg: The "comput_jet_Id" function will break if jet_type is "FatJet", because FatJet.neHEF and FatJet.neEmEF do not exist. Previously they will only be called if the analysis requires jetId==6. After this PR, the jet_selection function breaks once it's used for FatJets.

valsdav: let's follow up here: https://github.com/PocketCoffea/PocketCoffea/issues/427

PR #426, opened 2025-11-05
Proposal for a verbose debugger 

valsdav:
The PR proposes a debug logger object which gets propagated in the PocketCoffea processor and stores to files all the details of processing:

- computed weights and variations
   - check nan/inf/zero values
-  masks and preselections
-  calibrators applied
-  histogram filling details

Is this something that can be useful to track down problems on the user configuration side?

### DIFF
--- a/tests/test_full_configs/test_debug_logging/config.py
+++ b/tests/test_full_configs/test_debug_logging/config.py
@@ -1,0 +1,101 @@
+# Test configuration for super verbose debug logging feature
+from pocket_coffea.utils.configurator import Configurator
+from pocket_coffea.lib.cut_definition import Cut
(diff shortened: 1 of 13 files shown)


tests/test_full_configs/test_debug_logging/config.py
```diff
@@ -0,0 +1,101 @@
...
+            "year": ['2018']
+        }
+    },
+
+    workflow = BasicProcessor,
+    
+    # Workflow options with super debug enabled
+    workflow_options = {
+        "super_debug": {
+            "enabled": True,
```
felixzinn: does it make sense to introduce a flag in the processing script for the `enable` parameter? so that it can be easily deactivated without having to change the config?

felixzinn (commented):
is the general purpose of this to have a debugging/verification tool during testing
or a complete log of the full processing?

PR #436, opened 2025-12-02
Refactor download.py to handle files and metadata correctly

matteomalucchi:


matteomalucchi:
Hi @valsdav I used it now for the first time, but it was pretty useful to quickly download small datasets without having to pass by Rucio

PR #437, opened 2025-12-02
Export Columns with variations

felixzinn:
Goal: export events with varied weights and systematic variations

- for weight variations: export nominal and the varied weights (up/down)
- for systematic variations: export the varied event variables and the corresponding nominal weights (the shape of those is different to the nominal weight in the nominal variation due to cuts)

this results in the following structure

grafik

this is a first attempt, feedback is very welcome.
so far not implemented for export per chunk

FYI: @NinaHerfort


felixzinn:
oh, dont know how I missed that PR 🧐 i checked before 🤷🏼‍♂️ 
yes, thats essentially the same implementation
so we can close this one

PR #462, opened 2026-02-26
Add LPC (Fermilab) executor support with Dask and HTCondor integration

senadurgut:
This implementation enables one to run analysis using pocket-coffea config files on Fermilab's LPC cluster with either Dask (interactive) or HTCondor (batch) submission.


valsdav:
Hi @senadurgut, should we merge this?

SohamBhattacharya:
Hi, just wanted to bump this and add that being able to run on LPC would be be a **significant** upgrade! :)

PR #475, opened 2026-03-15
Add MuonsRochesterCalibrator for Run 2 UL muon momentum corrections

piosifid:
This PR adds support for Rochester momentum corrections for Run 2 UL muon collections, following the same pattern as the existing legacy calibrators.

**What was added:**

1. **MuonsRochesterCalibrator** in **pocket_coffea/lib/calibrators/legacy/legacy_calibrators.py**
2. **roccor_wrapper.py** — Python ctypes interface to the compiled RoccoR shared library
3. **Rochester correction parameters** added to l**epton_scale_factors.yaml under muon_sf.rochester**
4. **RoccoR source files and UL txt** correction files in **data/roccor/**
5. Test **test_muons_rochester_calibrator** in tests/test_calibrators.py

**How it works:**

- Corrects Muon.pt directly (not a weight)
- Data → kScaleDT
- MC → kSpreadMC where gen-matched muon is available, kSmearMC otherwise
- Provides nominal + muon_roccorUp / muon_roccorDown variations from stat replicas
- Automatically disabled for Run 3 years via the apply flag in the yaml

**Implementation approach:**

The official CMS RoccoR.cc is compiled once into RoccoR.so with no external dependencies
Avoids reimplementing the RoccoR logic — guarantees correctness
Compiled on lxplus-el9 (AlmaLinux 9, x86_64), verified to load inside the PocketCoffea apptainer
If preferred, a compilation step can be added to the apptainer build: g++ -O2 -shared -fPIC -std=c++17 -o RoccoR.so RoccoR_wrap.cc

**Notes:**

- UL txt correction files are included in data/roccor/ for convenience but can also be downloaded from the official source at gitlab.cern.ch/akhukhun/roccor
- The txt files and .so placement in the repo is open for discussion


pocket_coffea/parameters/roccor/RoccoR.cc
valsdav, line 1: Why this is under the "parameters" folder? I would move this to lib

piosifid: Moved the C++ source files (RoccoR.cc, RoccoR.h, RoccoR_standalone.h, RoccoR_wrap.cc) to lib/calibrators/legacy/. The runtime files (RoccoR.so and the UL txt correction files) stay in parameters/roccor/ since that's where they're loaded from at runtime.

PR #488, opened 2026-04-14
Make jet regression work for general user

andreypz:
Here are some minor modifications which are needed for regression to work in the following way:
* A user needs to define a new jet collection and assign it to "Jet". This can be done in user config, no need to modify the pocket-coffea base code parameters.
```
jets_calibration:
  collection:
    2022_postEE:
      #AK4PFPuppi: "Jet" 
      AK4PFPuppi: null
      AK4PFPuppiPNetRegression: "Jet" 
```
Here, the line ` AK4PFPuppi: null` is needed because in the default parameters the `AK4PFPuppi: "Jet"` is used. We can not  have two collections using the same object - there will be some collision. The modifications in the code basically allow for this and skip a collection if it is `None` (==`null` in yaml.)

In this way the regression will be applied to all jets (regardless of the tagger scores). There is no need to modify the processor code (the name of the jets is the same). In order to run without regression a user can just uncomment the `AK4PFPuppi: "Jet"` line and comment out the other two lines in their config.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -310,6 +310,39 @@ jets_calibration:
 Note that there are two versions of regression are available in PNet: with and without neutrinos used in the training.  
 In the example above, the default `jets` configuration is overwritten to assign the `Jet` collection to the `AK4PFPuppiPNetRegression` tag, and to activate the pt regression for data and MC for that tag. Note the line `AK4PFPuppi: null` -- it is needed to remove the association of the `AK4PFPuppi` to the `Jet`, which is default pocket-coffea setting.
 
+However, this is not all. We also need to apply JEC and JER on the regressed jets. Ideally, dedicated corrections should be applied to those, but these are not yet approved be JETMET group (time of writing: April 2026). Insted, one can apply the standard JEC/JER and the corresponding uncertainties. Foe these we need a few extra lines in the config:
+
+```yaml  
+jets_calibration: 
+  collection_name_alias:  # Alias for JEC/JER corrections and syst
+	2022_preEE:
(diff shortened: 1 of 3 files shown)


andreypz:
HI @valsdav , I think this PR can be merged. It should not affect anyone who is not using jet regression, and probably not even those who do use them. 
I think this enables anyone to use a version of the regression out of the box, simply re-defining config files, as written in the documentation.

valsdav:
Thanks Andreyz, merging this

valsdav:
Hi! The integration tests are failing with :

```
FAILED tests/test_calibrators.py::test_jets_calibrator - omegaconf.errors.ConfigKeyError: Missing key AK4CorrT1METJetchs
    full_key: jets_calibration.apply_pt_regr_MC.2018.AK4CorrT1METJetchs
```

I will have a look tomorrow.

andreypz:
How can I run the tests locally? Never done it, can't figure out.

I did change a config [here](https://github.com/PocketCoffea/PocketCoffea/pull/488/files#diff-5305607db1a0c7660e50204dc466a823980e5f01e4cd4008ecbf70db74f72225L625-L680) but have not touched the 2018 parameters, so it's a bit confusing why it complains about it.

By the way, just noticed that I missed that line for for `2022_preEE`. Should add in
```
apply_pt_regr_MC:
    2022_preEE:
      AK4CorrT1METJetPuppi: False
```

Could also add these for 206-2018, but not sure why...

valsdav:
For the tests 
1) enter apptainer
2) export PYTHONPATH="your local path to PocketCoffea folder"
2) pytest tests 

Ok it was just missing the key, I can make an hotfix for that. Thanks for checking!

PR #493, opened 2026-04-30
Hot fix for calibrator pytest

andreypz:
This may fix the failing test after #488 .
However, I could not fully confirm it, since the tests still fail due to xrootd error for some file.


valsdav:
Thanks! indeed it was needed also in the other years

PR #495, opened 2026-05-09
Fix type alias lookup order in JetsCalibrator to support pt regression

matteomalucchi:
- Moves the `collection_name_alias` resolution block before the pt regression check in `JetsCalibrator.__call__`
- Ensures `jet_type_alias` is defined before it is used when evaluating `apply_pt_regr_MC` / `apply_pt_regr_Data` for a given jet type and year
- Previously, if a jet type had a collection name alias, the pt regression logic would incorrectly look up the uncorrected `jet_type` key instead of the alias

### DIFF
--- a/pocket_coffea/lib/calibrators/common/common.py
+++ b/pocket_coffea/lib/calibrators/common/common.py
@@ -47,11 +47,19 @@ class JetsCalibrator(Calibrator):
             #print("Initializing:", jet_type, jet_coll_name)
             if jet_coll_name==None:
                 continue
-
+            
+            # Define the key name to get the jet corrections
+            if ( ("collection_name_alias" in self.jet_calib_param) and 
+                (self.year in self.jet_calib_param.collection_name_alias) and
+                (jet_type in self.jet_calib_param.collection_name_alias[self.year])):
+                jet_type_alias = self.jet_calib_param.collection_name_alias[self.year][jet_type]
+            else:
+                jet_type_alias=jet_type
+            
             # Check if the pt regression is requested. If so - aplly it and continue with JECs
-            if ((self.isMC and self.jet_calib_param.apply_pt_regr_MC[self.year][jet_type]) 
+            if ((self.isMC and self.jet_calib_param.apply_pt_regr_MC[self.year][jet_type_alias]) 
                     or
-               (not self.isMC and self.jet_calib_param.apply_pt_regr_Data[self.year][jet_type])):
+               (not self.isMC and self.jet_calib_param.apply_pt_regr_Data[self.year][jet_type_alias])):
                 # Get the regression parameters by collection if they are present
                 regression_params = OmegaConf.select(self.params,
                                                      "object_preselection." + jet_coll_name + ".regression")
@@ -67,13 +75,6 @@ class JetsCalibrator(Calibrator):
                 events[jet_coll_name] = jets_regressed
 
                 
-            # Define the key name to get the jet corrections
-            if ( ("collection_name_alias" in self.jet_calib_param) and 
-                (self.year in self.jet_calib_param.collection_name_alias) and
-                (jet_type in self.jet_calib_param.collection_name_alias[self.year])):
-                jet_type_alias = self.jet_calib_param.collection_name_alias[self.year][jet_type]
-            else:
-                jet_type_alias=jet_type
                 
             # Check if the collection is enables in the parameters
             # print("Working on ", jet_type, jet_coll_name )


andreypz:
Hi @matteomalucchi ,
Unfortunately, this does not work with my setup of regression.  I think this depends on configuration and the code in the workflow. I have described my version of the config files for regression in the docs: https://pocketcoffea.readthedocs.io/en/latest/recipes.html#jet-energy-regression
Could you please test if this works for you?
The idea is to replace **all** jets with regressed ones. This should be a default and simplest implementation (does not require changes in the user's workflow code). 
My config file for this to work looks like this: [link to gitlab](https://gitlab.cern.ch/cms-analysis/hig/vhcc-run3/VHccPoCo/-/blob/5c0104ec013999d4e119f8d46b58018ab49ed045/params/jet_regression.yaml). As you can see, I have `collection_name_alias` for `AK4PFPuppiPNetRegression` set to `AK4PFPuppi`. But the regression flag for `AK4PFPuppi` jets is set to `False` in PocketCoffea parameter `apply_pt_regr_MC` [here](https://github.com/PocketCoffea/PocketCoffea/blob/ae00f71bb9649995378653c5710c62b394c23ba9/pocket_coffea/parameters/jets_calibration.yaml#L603) and according to your changes the regression will not be applied. 

We either need to revert this change, or update the documentation with correct instructions.

matteomalucchi:
HI @andreypz, 
I think I see where the problem comes from. I think we shouldn't set
```

  collection_name_alias:  # Alias for JEC/JER corrections and syst
    2022_preEE:
      AK4PFPuppiPNetRegression: "AK4PFPuppi"
```
since the corrections are different in general. 
At the moment the corrections for the regression are not centrally on `/cvmfs` and they are only on `/eos` (here is the link to the [documentation from JME](https://cms-jerc.web.cern.ch/ExpJEC/#jec-for-pnet-and-upart-regressed-jets)). Probably the best plan would be to copy the files directly in the central repository and reference them in the `jets_calibration.yaml`. 
Until this is done I think the best way is to not use the same name alias as the `AK4PFPuppi`, but to define the `AK4PFPuppiPNetRegression` as a separate collection and share **only** the path to the JERC files with `AK4PFPuppi`, like this
```
jets_calibration:
  collection:
    2022_preEE:
      #AK4PFPuppi: "Jet"
      AK4PFPuppi: null
      AK4PFPuppiPNetRegression: "Jet"

default_jets_calibration:
  factory_config_clib: 
     AK4PFPuppiPNetRegression: "${default_jets_calibration.factory_config_clib.AK4PFPuppi}" 
```

In [my configuration](https://github.com/matteomalucchi/AnalysisConfigs/blob/main/configs/HH4b_common/params/jets_calibration_regression_json.yaml) I have added the json files for the regression, so I put the path directly o those files.

Let me know what you think.

andreypz:
Hi @matteomalucchi , I don't think there is a problem with my `collection_name_alias` definition. Rather, it is a feature!
Indeed I did not use the dedicate JERCs for the regressed jets. Instead I applied the regular Puppi JERCs on them. And this `alias` is designed to do that (as far as I understand). The change in this PR breaks it.
Moreover, I believe, even for the config you pointed to in your analysis, I think you do not need the change in this PR...
Since you do not have the `collection_name_alias` fields, your `jet_type_alias` will by just `jet_type` anyway. So what problem/error did you encounter with your config?

@valsdav if Matteo agrees with my arguments, I think this PR should be reverted.

By the way, thanks for pointing to the latest JERCs for regressed jets (they were not available last time I checked). I think I made them to work in my setup as well. I could update the documentation using those correction in my next PR. 
Are you the one producing these corrections fro JetMet? If so, could you push them to cvmfs as well?  The eos area is not as convenient as cvmfs.

matteomalucchi:
Hi @andreypz, my idea for  `collection_name_alias` was to have **exactly** the same corrections as the collection the dictionary points to. This would include the application of the regression, the variations and which JERC to apply. So I think in your case it would be best to not use the alias between the regressed pt and standard puppi jet pt. 

Regarding the corrections, I am involved in the derivation but I'm not the one producing the json files. I think that the long term plan for JME is to include them on `cvmfs`, but I'm not sure. I will ask

andreypz:
Ok, I see what you mean. If that's the intention, then I understand.
However, wouldn't it be more versatile to separate the regression and JERC application? Then one could setup different combinations, like I did with regressed jets and Puppi JERCs. For this case, you would only have set the `apply_pt_regr` for each jet collection explicitly.

matteomalucchi:
Since the alias also sets the same variations, I would keep it such that the alias collection has exactly the same pT (regression included). In your case you should just add this and it should be equivalent I think
```
default_jets_calibration:
  factory_config_clib: 
     AK4PFPuppiPNetRegression: "${default_jets_calibration.factory_config_clib.AK4PFPuppi}" 
```

PR #498, opened 2026-05-19
Minor fixes for PNetRegression

andreypz:
Here is a list of changes:
* The most important change here is an addition of two lines in `pocket_coffea/lib/jets.py`
```python
    if 'AK4PFPuppiPNetRegression' in jet_algo:
        jet_algo='AK4PFPuppi'
````
This is a hack needed for Nanov15 jet id for regressed jets to work. 
However, one should come up with a better solution in the future, because there is still going to be a problem with other custom jet collections.
* Changes in documentation for PNet regression
* Switching default flags in `parameters/jets_calibration.yaml`. These flags do not affect anyone who does not use the regression. But for those who do, it is more convenient to have them switched on in default config, so that it is not needed to add them in user config.
* Some missing flags in `lepton_scale_factors.yaml` (for trigger SFs), though they are probably not used by anyone. Maybe should delete them in the future?
* Change in `utils/rucio,py` to ckeck if `rse` field actually exists. We found a site where it did not.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -495,7 +495,7 @@ jets_calibration:
 
   # Path to JERCS at CERN EOS:
   path_to_JERC_PNetReg: '/eos/cms/store/group/phys_jetmet/ExpJEC/json_files_Reg/'
-  # If running outside CERN without access to EOS - copy the files to your cluster and use the appropriate the path.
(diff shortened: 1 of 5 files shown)


andreypz:
@valsdav this is ready to be merged, please take a look

PR #499, opened 2026-05-25
Improvements for jobs runner

valsdav:
- Documentation
- Resubmit-jobs with site blocklist for replacing input files
- `max-events-per-job` split by sample

### DIFF
--- a/pocket_coffea/scripts/check_jobs.py
+++ b/pocket_coffea/scripts/check_jobs.py
@@ -27,7 +27,8 @@ import time
 import re
 import cloudpickle
 from copy import deepcopy
-from pocket_coffea.utils.rucio import get_xrootd_sites_map
+from pocket_coffea.utils.rucio import get_xrootd_sites_map, get_rucio_client
+from pocket_coffea.utils.site_rewrite import _query_replicas
 from collections import Counter
 
 queues = [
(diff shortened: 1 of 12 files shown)


pocket_coffea/scripts/check_jobs.py
```diff
@@ -96,16 +97,15 @@ def check_jobs_logs(jobs_folder):
...
     if filepath.startswith("root:/"):
         rootpref = filepath.split("/store/")[0]
         file = "/store/"+filepath.split("/store/")[1]
     else:
         rootpref = None
         file = filepath
-    
-    command = f'/cvmfs/cms.cern.ch/common/dasgoclient -query="site file={file}"' 
-    sites = os.popen(command,'r').read().split()
+
```
valsdav, line 107: @mondalspandan I'm replacing dasgoclient with the internal rucio tool here. I think it's more direct.

mondalspandan: Thanks for the heads up, Davide! 

On a related note, I have been making some changes here again, and I was wondering why you chose .running/.idle status files on disk instead of periodic `condor_q` queries inside `check_jobs`. Is there an obvious problem with the latter?

valsdav: ehi :) 
Well usually keep querying the condor scheduler can create problems I think. Checking the existance of an empty file on disk it seems more efficient to me. The mechanism is quite robust in my experience, especially with the various resubmit options we have now.

mondalspandan: I see. The problem with .status files that I faced is that I hit the upper limit of max allowed files in each afs directory while trying out a job-splitting scheme for timeout jobs. But I agree in general it's a robust way to keep track.

Issue #511, opened 2026-06-10
Update plotting code to work with mplhep>=1.0.0


felixzinn:
plotting code in PocketCoffea does not produce correctly styled plots (e.g. labels, lumi) with the new major mplhep v1 release (https://github.com/scikit-hep/mplhep/releases)
some functions have been renamed, new functionality has been introduced

PR #513, opened 2026-06-10
update plotting utilities #511 

felixzinn:
added some basic tests to compare plots when updating the plotting code
to start it just checks if the plots are created, maybe we could use https://github.com/matplotlib/pytest-mpl as is also done by mplhep package


felixzinn:
the PlotManager has an argument `toplabel` which by default is the lumi for the year
https://github.com/PocketCoffea/PocketCoffea/blob/2cc66c3ad1977dba272a0758213c3d4c1216a877/pocket_coffea/utils/plot_utils.py#L214-L217

in the old mplhep version this was then set with `mplhep.cms.lumitext`
https://github.com/PocketCoffea/PocketCoffea/blob/2cc66c3ad1977dba272a0758213c3d4c1216a877/pocket_coffea/utils/plot_utils.py#L1038-L1043

which however does not exist in the current version of `mplhep>=1.2.0`
most of the styling is done with `mplhep.cms.label` (https://mplhep.readthedocs.io/en/latest/api/#mplhep.label.exp_label)
with convenient arguments to specify (from the documentation)
- `text`: after CMS e.g. Preliminary
- `data`: if data is used, will prepend "Simulation" to the CMS label text
- `year`, `lumi`, `com`

which would automatically format the CMS label and the lumi/year/com information in the top right
it is possible to provide custom labels with
- `llabel`: Manual override for left-hand label text. Overrides "data" and "text" parameters.
- `rlabel`: Manual override for right-hand label text. Overrides "year", "lumi", "com" parameters.

what is called `toplabel` so far is probably what should be `rlabel` in the new mplhep version
_but_ it needs to be passed to the PlotManager
https://github.com/PocketCoffea/PocketCoffea/blob/2cc66c3ad1977dba272a0758213c3d4c1216a877/pocket_coffea/utils/plot_utils.py#L147-L169

and _not_ to the style config. Unlike the custom cms_label which can be specified in the plotting_style yaml section and is the text behind the CMS text
https://github.com/PocketCoffea/PocketCoffea/blob/2cc66c3ad1977dba272a0758213c3d4c1216a877/pocket_coffea/utils/plot_utils.py#L1016-L1022

so what do you think about providing the option to specify `llabel` and `rlabel` in the style config (yaml file) and bool values to specify if `year`, `lumi`, `com` should be shown in the plot

felixzinn:
`toplabel` is also not passed in the plotting script provided by PocketCoffea
https://github.com/PocketCoffea/PocketCoffea/blob/2cc66c3ad1977dba272a0758213c3d4c1216a877/pocket_coffea/scripts/plot/make_plots.py#L141-L157

felixzinn:
mplhep provides a way to save multiple versions of a figure with different labels, e.g. Work in Progress, Preliminary, ...
https://mplhep.readthedocs.io/en/latest/api/#mplhep.label.savelabels
should we provide this functionality? could be an option for the plotting script

tests/test_plotting.py
```diff
@@ -0,0 +1,379 @@
...
+        if with_ratio:
+            assert len(saved_files) == 2
+        else:
+            assert len(saved_files) == 1
+
+        for name in saved_files:
+            path = output_dir / name
+            assert path.exists(), f"expected plot file {path} was not created"
+
+    # def test_plot_cutflow_from_output(self, tmp_path: Path, coffea_output: dict):
```
felixzinn: this test fails with the current package requirements
I assume it is coffea 0.7.29
the cutflow dictionary is loaded and the values are of type `np.int64` which are not an instance of `int` so this line
https://github.com/PocketCoffea/PocketCoffea/blob/5743308a3b53b3e46fb551b5ffdd4a8a2e6a3028/pocket_coffea/utils/cutflow_utils.py#L113
fails.

I tested it with the coffea2026 and there it works
I am not sure what is happening here and why it has not been noticed so far.
So once PocketCoffea is moved to the new coffea versions we can include this test

tests/test_plotting.py
```diff
@@ -0,0 +1,379 @@
...
+            for category in shape_object.categories:
+                plot_path = (
+                    plot_mngr.plot_dir
+                    / category
+                    / f"{shape_object.name}_{category}.png"
+                )
+                assert plot_path.exists(), f"Plot {plot_path} was not created"
+
+
+# from pocket_coffea.utils.plot_functions import plot_shapes_comparison
```
felixzinn: this function is not used in any scripts but just in two jupyter notebooks
- https://github.com/PocketCoffea/PocketCoffea/blob/main/notebooks/BtagSF.ipynb
- https://github.com/PocketCoffea/PocketCoffea/blob/main/notebooks/JetPUId.ipynb
so not sure if we need a test for this?

PR #515, opened 2026-06-15
Fix tags for all corrections from /cvmfs/cms-griddata.cern.ch/cat/metadata/

anigamova:
The preliminary branch was removed, fixing .json file location

### DIFF
--- a/pocket_coffea/parameters/jet_scale_factors.yaml
+++ b/pocket_coffea/parameters/jet_scale_factors.yaml
@@ -14,98 +14,108 @@ jet_scale_factors:
     btagSF:
         # DeepJet AK4 tagger shape SF
         '2016_PreVFP':
-            file: ${cvmfs:Run2-2016postVFP-UL-NanoAODv9,BTV,btagging.json.gz}
+            file: ${cvmfs:Run2-2016postVFP-UL-NanoAODv9,BTV,btagging.json.gz,2025-08-19}
             name: "deepJet_shape"
         '2016_PostVFP':
-            file: ${cvmfs:Run2-2016preVFP-UL-NanoAODv9,BTV,btagging.json.gz}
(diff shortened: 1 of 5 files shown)


valsdav:
Hi @anigamova. I saw that `prelim` was gone, but latest contains this file.   Would you prefer fixing the tag? Usually we stay on latest in the defaults of the framework

anigamova:
Hi Davide, I think for the analysis we should always be pointing to the latest tagged version, but not the `latest` itself. I see that for [jet calibration](https://github.com/PocketCoffea/PocketCoffea/blob/main/pocket_coffea/parameters/jets_calibration.yaml) you use tags not the latest it self, which I think is safer but it adds a bit more maintenance tasks. In general, I think it would be great to make it consistent.

valsdav:
For the JECs, we added the specific tags as the new latest tag requires some code changes ongoing. 

Let's say that using explicit tag is more stable, but then we may miss updates in the central defaults. 
The original philosophy was that analyzers can freeze tags in their configuration while getting always the latest working from the framework defaults.   

What do you think about this?

anigamova:
Thanks Davide! I think it makes sense in terms of minimizing the effort. However, it also means that we have to make sure that the changes introduced to the corrections implementations are backwards compatible. So that the users using the dated tags can use newer version of pocket-coffea.

andreypz:
Hi @anigamova @valsdav 
I think there is a mistake in this PR for Run3 (2022 and 2023 in particular) https://github.com/PocketCoffea/PocketCoffea/pull/515/changes#diff-5305607db1a0c7660e50204dc466a823980e5f01e4cd4008ecbf70db74f72225L49-L50

The tag `2026-06-05` is incompatible with ` jec_mc: Summer22_22Sep2023_V3_MC`. One has to use earlier tag, I think.

Probably other eras are problematic too,, but I did not test

anigamova:
Hi @andreypz, I will make another pass, thanks!

PR #508, opened 2026-06-05
Jerc unc changes

phnattla:
With new JER SF for 2024 and 2025 the format was changed [1], necessitating updated parameters and JERC uncertainty handling. Now, the implementation also has to be different for v9 and v15. 
Tested for UL2018 nanov9 and 2024 v15.

@valsdav

[1] https://cms-talk.web.cern.ch/t/new-jer-smearing-inputs-available-for-2024-and-2025/145723

PR #519, opened 2026-07-02
Correct JEC correction sets

anigamova:


andreypz:
Is the code actually ready to move to new JERC corrections? 
I thought this still needs to be tested, and this code is needed: https://github.com/PocketCoffea/PocketCoffea/pull/508

I think the idea of tagging the JERCs was to fix the tag to an older version, `2026-04-13` for 2022/20223

anigamova:
As I understood from Davide it is about to get merged. If it does not happen we can revert indeed

valsdav:
HI! The fix for the latest JER have been merged. Merging this to update the tags

Issue #521, opened 2026-07-06
Claude Fable review of PocketCoffea


valsdav:
I have tested Claude Fable model for a full review of the PocketCoffea `latest` version. 
This is the full summary of the analysis. A series of PRs will follow with fixes and improvements. 

**No big issues identified in the nominal loop**. Some moderate issues in the systematic variations (i.e. MET calibration for JECs variation).  For the rest, mainly possible improvements and cleanups.

---
# PocketCoffea — Full Package Review

- **Date:** 2026-07-03
- **Code state:** `main @ 0b6e11b6` (~32k lines of Python in `pocket_coffea/`)
- **Method:** six parallel deep code-review passes over the subsystems (core pipeline;
  calibrators; weights/histograms; cuts/objects/categorization; executors/CLI;
  dataset/plotting/stat utils), each producing candidate findings with quoted evidence.
  Every high-severity finding was then **re-verified by hand against the source**, and
  runtime-behavior claims were reproduced inside the CI container
  (`gitlab-registry.cern.ch/cms-analysis/general/pocketcoffea:lxplus-el9-latest`,
  Python 3.11, coffea 0.7.29, awkward 1.10.5) with the local checkout on `PYTHONPATH`
  and a valid grid proxy. Performance claims are grounded in a `cProfile` run of a real
  analysis config.
- **Line numbers** refer to `main @ 0b6e11b6`.

---

## 1. Executive summary

The core processing pipeline — the calibration loop with per-variation snapshot/reset,
weight composition across categories and subsamples, skim genweight reconstruction, cut
caching — is **architecturally sound and correct on the paths the test suite
exercises**. All 80 tests pass in the CI container, including both skim modes end-to-end
against EOS. The review found no double counting, no up/down swaps, and no cross-chunk
state leakage in the main MC histogramming path.

The defects cluster in three places:

1. **The type-1 MET calibrator**, whose interaction with jet variations and with Run-2
   defaults is quantitatively wrong (Findings A1, A2).
2. **Less-exercised branches** — resubmission tooling, per-subsample columns, datacard
   rate formatting, dataset-query metadata, flavour-split gen leptons, non-standard
   trigger names — where bugs corrupt output **silently** (Findings A3–A8) or crash a
   documented feature outright (Findings B1–B5).
3. **Systematic re-creation of correction evaluators** per chunk and per variation,
   which profiling shows is the dominant avoidable CPU cost (Findings P1–P4).

[...]

PR #522, opened 2026-07-07
Fix trigger-name prefix stripping (lstrip char-set bug)

valsdav:
`str.lstrip("HLT_")` / `lstrip("L1_")` strip a *character set*, not a prefix, so trigger names starting with H/L/T after the prefix were mangled, e.g. `HLT_TkMu50` -> `kMu50` and `HLT_TripleMu_12_10_5` -> `ripleMu_12_10_5`. `apply_trigger_mask` then silently skipped the mangled (unknown) name, dropping it from the trigger OR and losing events with no error.

- Add `remove_trigger_prefix()` (genuine prefix removal) and route all four sites through it (triggers.py x3 + cut_functions.py custom-trigger helpers).
- Log a de-duplicated WARNING when a requested trigger is absent from the file, so a genuinely-missing or mistyped path is surfaced instead of vanishing.
- Add offline unit tests (synthetic `.HLT` record) covering the prefix removal, the OR, the missing-trigger warning, and a `get_trigger_mask_byprimarydataset` regression on `HLT_TkMu50`.

Behaviour change: configs whose trigger names were being silently dropped will now include those paths in the OR (that is the fix). The 2017 Ele32 special-case guard is intentionally left unchanged pending analyst confirmation.

### DIFF
--- a/pocket_coffea/lib/cut_functions.py
+++ b/pocket_coffea/lib/cut_functions.py
@@ -1,6 +1,6 @@
 import awkward as ak
 from .cut_definition import Cut
-from .triggers import get_trigger_mask_byprimarydataset,  apply_trigger_mask
+from .triggers import get_trigger_mask_byprimarydataset,  apply_trigger_mask, remove_trigger_prefix
 import correctionlib
 import numpy as np
 from coffea.lumi_tools import LumiMask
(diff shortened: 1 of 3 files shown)


valsdav:

PR #523, opened 2026-07-07
Fix getGenLeptons flavour split misalignment

valsdav:
The Muon/Electron branches indexed the filtered, pt-sorted `leptons` collection with a boolean mask built on the full `GenPart` (`particles`). The two have different per-event lengths, so the mask is misaligned — in awkward 1.x this does not raise but silently returns leptons scrambled across event boundaries, making the flavour-split paths return physically meaningless objects.

- Build the flavour mask from `leptons` itself (it retains the `pdgId` field).
- Fix the unsupported-flavour branch: `raise ValueError(...)` instead of raising a tuple (which itself raised TypeError).
- Add an offline unit test (synthetic GenPart) asserting the split is aligned and correctly partitions "Both".


valsdav:

PR #524, opened 2026-07-07
Apply minpt in nElectron/nMuon; include collection in Cut identity

valsdav:
nElectron/nMuon accepted a `minpt` argument, baked it into the cut name, but never applied it — the mask was just `events.nXGood >= N`. Count objects above the threshold instead. Behaviour change: configs using `get_nElectron(N, minpt>0)` (or nMuon) now honour the pt cut; minpt=0 is unchanged.

Cut.__hash__ omitted `collection`, so two cuts identical except for the collection they act on (e.g. "events" vs "Jet", producing masks of different dimensionality) hashed/compared equal and were deduplicated to one in StandardSelection. Add `collection` to the identity, and serialize params with sort_keys=True (order-independent) and default=str (tolerate numpy scalars).

Adds offline unit tests for the minpt threshold and the collection-aware, param-order-independent, numpy-safe Cut identity. Verified test_categorization and test_cartesian_categorization still pass (cut-id change does not alter output).

### DIFF
--- a/pocket_coffea/lib/cut_definition.py
+++ b/pocket_coffea/lib/cut_definition.py
@@ -32,8 +32,22 @@ class Cut:
         )
 
     def __hash__(self):
-        '''The Cut is unique by its name, the  function, and the dict of parameters.'''
-        return hash((self.name, json.dumps(self.params), self.function))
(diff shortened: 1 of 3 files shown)


valsdav:

PR #525, opened 2026-07-07
Fix CartesianSelection serialize/template_mask and StandardSelection

valsdav:
- CartesianSelection.serialize() called self.common_cats.serialize(), but when no common categories are given self.common_cats is a plain {} (no serialize()). Since the Configurator serializes categories at every run start, a CartesianSelection built from only multicuts crashed with AttributeError at config-save time. Guard it (common_categories -> None when absent).
- CartesianSelection.template_mask referenced self.ready/self.storage/self.ismulti_dim, none of which are ever set on that class -> AttributeError if accessed. Delegate to the common StandardSelection (or None), matching how the class actually stores masks.
- StandardSelection.template_mask used the misspelled self.ismulti_dim -> is_multidim.
- StandardSelection.__init__ error path formatted an undefined `cut` (loop var is `c`) -> NameError instead of the intended message. Use `c`.

Adds an offline test building a CartesianSelection with only multicuts and asserting serialize()/template_mask no longer raise, plus the corrected error message.

### DIFF
--- a/pocket_coffea/lib/categorization.py
+++ b/pocket_coffea/lib/categorization.py
@@ -221,7 +221,7 @@ class StandardSelection:
             self.categories[k] = []
             for c in cuts:
                 if not isinstance(c, Cut):
-                    raise Exception(f"The cut is not defined as a Cut object: {cut}")
+                    raise Exception(f"The cut is not defined as a Cut object: {c}")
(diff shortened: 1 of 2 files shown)


valsdav:

PR #527, opened 2026-07-07
Fix shared metadata dict corrupting das_names across selected datasets

valsdav:
do_select appended the *same* `metadata` dict object once per dataset when a (wildcard) query selected several datasets with one metadata. do_save then mutated that shared dict in place (das_names, sample), so every dataset in the group ended up with the last dataset's das_names — both in the *_replicas file and, via the same reference, in the main dataset definition.

- Deep-copy the metadata per dataset in do_select (root cause), and defensively deep-copy again when materialising it in do_save so no shared object is mutated.
- generate_default_metadata: warn instead of silently defaulting xsec=1.0 on lookup failure, since that wrong normalization is written into the dataset definition.

Adds an offline test (rucio stubbed when unavailable) asserting do_select stores independent per-dataset metadata dicts.

Note: other dataset-discovery issues (UL16 PreVFP/PostVFP year detection, the --check/--download code paths, rucio partial_allowed) are left as separate follow-ups.

### DIFF
--- a/pocket_coffea/scripts/dataset/dataset_query.py
+++ b/pocket_coffea/scripts/dataset/dataset_query.py
@@ -243,6 +243,12 @@ Some basic commands:
         try:
             xsec = self.extract_xsec_from_dataset_name(dataset)
         except Exception as e:
+            # Falling back to 1.0 silently writes a wrong MC normalization into the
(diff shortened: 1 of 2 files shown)


valsdav:

PR #529, opened 2026-07-07
Apply dr/pt cut before greedy assignment in object_matching

valsdav:
object_matching sorted candidate pairs by deltaR and let the greedy matcher (get_matching_pairs_indices) consume each object the first time it appeared, then applied the deltaR+pt mask only afterwards. A pair with small deltaR but failing the dpt_max cut therefore consumed its objects and blocked an otherwise-valid, larger-deltaR pair that passed both cuts, leaving objects unmatched. Pure-deltaR use was already correct (failing pairs sort last); the bug bit whenever dpt_max was given.

Filter the pairs by the combined mask *before* the greedy assignment so cut-failing pairs never consume objects. Pure-deltaR results are unchanged.

Adds offline tests (coffea vector behavior) for both the dpt-blocking case and an unchanged pure-deltaR 1-1 match.

Note: the leptons.py electron-smear correlation and get_dilepton ordering items from the review are moved to the calibrators PR, which already edits leptons.py.

### DIFF
--- a/pocket_coffea/lib/deltaR_matching.py
+++ b/pocket_coffea/lib/deltaR_matching.py
@@ -100,16 +100,23 @@ def object_matching(obj, obj2, dr_min, dpt_max=None, return_indices=False):
         maskPt = deltaPt < dpt_max
         maskDR = maskDR & maskPt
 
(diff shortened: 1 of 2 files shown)


valsdav:

PR #530, opened 2026-07-07
Fix condor job.sh generation and recreate-jobs crashes

valsdav:
- run_with_retries used `$1` for the flag-file names, but inside a bash function `$1` is the function's first argument (the command string), not the job id. On copy exhaustion it wrote garbage flag files and never created job_<id>.failed, so check-jobs saw the job as running forever. Capture the id once as JOBID="$1" and use $JOBID throughout (incl. inside the function).
- Split-by-category: cd into the job-local `output/` scratch dir (not the shared final dir, where split-output finds no input and the glob picks up other jobs' outputs), and exit-check cd/split/copy so a failure marks .failed instead of falling through to .done.
- --recreate-jobs <explicit ids> raised NameError: failedjobs/runningjobs/idlejobs were defined only in the `auto` branch but used for explicit resubmissions. Compute them unconditionally.
- voms-proxy run option: `if vomsproxy := d.get(...) is not None` bound the boolean, not the path, so a supplied proxy set x509_path=True. Parenthesise the walrus. (both executors_base and executors_manual_jobs)

Extracts build_job_script() so the wrapper is unit-tested without a live submission; adds tests for the flag-file naming, non-split copy, and exit-checked split path. Existing manual-jobs splitting tests still pass.

### DIFF
--- a/pocket_coffea/executors/executors_base.py
+++ b/pocket_coffea/executors/executors_base.py
@@ -21,7 +21,7 @@ class ExecutorFactoryABC(ABC):
 
     def setup_proxyfile(self):
         if self.run_options['ignore-grid-certificate']: return
-        if vomsproxy:=self.run_options.get('voms-proxy', None) is not None:
+        if (vomsproxy := self.run_options.get('voms-proxy', None)) is not None:
              self.x509_path = vomsproxy
         else:
              _x509_localpath = get_proxy_path()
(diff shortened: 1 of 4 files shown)


mondalspandan:
Thanks for the fixes. I think we should deprecate `--recreate-jobs` entirely at this point. I can do that later if you prefer to merge this right now.

valsdav:
Do you think it is better to remove it because it is duplicated with the features in `check-jobs`? 
I saw that we have the job recreation path in 2 places, but I'm still trying to understand the different use-cases

mondalspandan:
Yes, exactly. It's duplicated and it's a bit painful to maintain both. Check-jobs basically does everything this script does plus some more stuff.

valsdav:
I agree, I have implemented a refactoring in #541

PR #531, opened 2026-07-07
Fix local hadd crash and merge_outputs robustness

valsdav:
- hadd_skimmed_files.do_hadd referenced R (ROOT), but `import ROOT as R` was local to hadd_skimmed_files() only, so the module-level do_hadd used by Pool.map raised NameError; and its `except subprocess.CalledProcessError` never matched (the body uses ROOT, not subprocess), so any real ROOT failure crashed the whole pool. Import ROOT inside do_hadd and broaden the except.
- merge_outputs: initialise job_config so the "no input files and no -jc" branch prints its message and exits instead of raising NameError.
- merge_outputs: load the reference file once instead of re-deserialising it for every comparison (it was reloaded N-1 times).
- merge_outputs: the type-mismatch report indexed type_mismatches (len N-1) with enumerate(inputfiles) (len N) -> IndexError on a genuine mismatch; zip against inputfiles[1:] instead.

Adds offline tests (ROOT stubbed) for do_hadd's R binding, the no-input exit, and the load-once behaviour.

Note: the merge auto-postprocess double-rescale guard and the runner adapt_chunksize / --process-separately items from the review are left as separate follow-ups.

### DIFF
--- a/pocket_coffea/scripts/hadd_skimmed_files.py
+++ b/pocket_coffea/scripts/hadd_skimmed_files.py
@@ -118,6 +118,11 @@ def _write_do_hadd_job_splitbyfile_script(path="do_hadd_job_splitbyfile.py"):
 
 
 def do_hadd(group, overwrite=False):
+    # ROOT is imported here (not at module scope) so this function works both in the
+    # local Pool.map path and when it is inherited by worker processes; the previous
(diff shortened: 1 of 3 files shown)


valsdav:

PR #532, opened 2026-07-07
Fixes for Histogram/Weights handling

valsdav:
- get_hist_axis_from_config passed underflow= to hist.axis.IntCategory, which does not accept it (verified against hist 2.9) -> any `intcat` axis crashed at histogram creation. Drop the unsupported kwarg.
- The fill loop iterated every category regardless of a histogram's only_categories, so excluded categories were filled into the (growth=False) category-axis overflow bin, silently contaminating flow-inclusive projections and wasting the masking/flatten work. Skip categories not in histo.only_categories.
- The weights_cache keyed the broadcast weight on (category, subsample, variation) only. Custom weights are a per-histogram dict and a collapsed 2D->1D mask depends on the histogram's collapse mode, so two histograms sharing that key got each other's weight. Key those cases per-histogram; the common path keeps the shared cache, so standard output is unchanged (test_new_weights baseline byte-identical).
- WeightLambda.compute wrapped a 1-element list return as the nominal weight instead of unwrapping out[0], which coffea then rejected as not-an-array.

Adds offline unit tests for the intcat axis and both WeightLambda return forms. Verified against full-config baselines: test_new_weights, test_custom_weights, test_custom_weights_on_data, test_shape_variations_and_weights_bysubsample, test_columns_export all pass unchanged.

### DIFF
--- a/pocket_coffea/lib/hist_manager.py
+++ b/pocket_coffea/lib/hist_manager.py
@@ -99,12 +99,13 @@ def get_hist_axis_from_config(ax: Axis):
             growth=ax.growth,
         )
     elif ax.type == "intcat":
+        # hist.axis.IntCategory has no `underflow` (categories have only an overflow
+        # "other" bin); passing underflow= raises TypeError, so it is dropped here.
         return hist.axis.IntCategory(
             ax.bins,
             name=ax.name,
             label=ax.label,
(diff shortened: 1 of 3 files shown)


valsdav:

PR #533, opened 2026-07-07
Fix/weights subsample followup

valsdav:
By-subsample weights were dropped in two places and two latent WeightsManager APIs were broken:
  
  - Exported columns used the full-sample weight only. columns_manager now threads the
    subsample name into WeightsManager.get_weight(subsample=...), which already folds the
    by-subsample factor; base.py supplies "sample__subsample". Fixes MC and data columns.
  - The data histogram fill branch never applied by-subsample weights. hist_manager now
    preloads weights_sub for data too (a no-op for unweighted subsamples, which return
    ones) and multiplies it in the data branch. Verified safe: the configurator builds a
    by_subsample entry for every subsample of every sample (MC and data), so
    get_weight_only_subsample cannot raise.
  - get_weight's subsample block did `overall_weight *= subsample_weight`, mutating
    coffea's internally-cached _weightsIncl array in place (.weight() returns it by
    reference).  **N.B.: no code in the current main branch was using this code path**. 
    Once columns started calling get_weight(subsample=...) per subsample, the
    first subsample's *= corrupted the full-sample weight for the next call. Switch the
    five *= to non-mutating `overall_weight = overall_weight * ...`. This was latent
    (nothing called get_weight with subsample=); get_weight_only_subsample already used *.
  - get_available_modifiers_bycategory combined sets with `+` (TypeError) and could fall
    through to an implicit None; rewritten with set-union. get_weight's subsample-only
    modifier branches left overall_weight None before `*=`; initialise from the inclusive
    nominal.
  - weights_run3 re-registered the existing sf_ele_trigger name (unimportable duplicate);
    re-export it instead.
  
  Tests: new full-config test_weights_data_subsamples configures a by-subsample weight on
  DATA (two same-selection subsamples, only one weighted) and asserts the 2x factor in both
  histograms and the exported weight column (this is what caught the get_weight mutation
  bug). Plus offline unit tests. test_columns_export, test_columns_export_parquet,
  test_custom_weights_on_data and test_new_weights pass unchanged.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1206,6 +1206,35 @@ files separately: the option is under development.
 :::
 
 
+### Exported weight columns
+
+Besides the `ColOut` collections you request, the columns output **automatically** includes
+the event weight, so the exported arrays can be used directly for weighted analysis or ML
+training without recomputing it:
+
+- **`weight`** — the nominal total event weight of that category: the product of the
+  inclusive weights, the by-category weights and, for a `sample__subsample` output, the
+  by-subsample weights. It is exactly the weight used to fill the histograms of the same
+  (sample/subsample, category), masked with the same category (and subsample) selection as
+  the exported columns.
+- **`weight_variation_{name}`** — *MC only*. One column per weight variation configured in
+  the `variations["weights"]` block (e.g. `weight_variation_sf_btagUp`,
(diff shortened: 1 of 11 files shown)


valsdav:

PR #534, opened 2026-07-07
Sort nominal jets consistently and remap jetIdx

valsdav:
The JetsCalibrator re-sorted jets by corrected pt only for the systematic variations, not for the nominal, so the leading-jet definition was inconsistent between nominal and shifts. Worse, consumers that look a jet up by a per-jet index (Electron/Muon.jetIdx, which point into the original NanoAOD order) read the *re-sorted* collection directly, so the closest-jet b-tag used for the promptMVA lepton selection and the mvaTTH BDT inputs was taken from the wrong jet whenever the collection had been re-sorted.

- JetsCalibrator now re-sorts the nominal AND the varied jet collections uniformly (the sort is hoisted out of apply_variation into a single post-step in calibrate), so the leading jet is defined the same way everywhere.
- The applied permutation is recorded as a field of the jet collection (jets.JET_SORTIDX_FIELD = "pocket_sortidx") instead of a separate top-level "Jet_sortidx" column. Storing it on the collection means the manager's reset restores the original jets (dropping the field) cleanly — the previous top-level column was never snapshotted (copy.copy raised) and leaked across variations;  It also lets the permutation travel with the jets into the xgboost evaluator, whose jets argument is events["Jet"].
- New helper jets.jets_in_original_order undoes the re-sorting (argsort of the stored permutation). leptons.lepton_selection_promptMVA and both xgboost_evaluator prepare_inputs use it before the jetIdx b-tag lookup; it is a no-op when the field is absent (sorting disabled / untouched collection).

Behavior: nominal jet ordering now follows the corrected pt (was NanoAOD/raw-pt order). In practice the JEC preserves the ranking almost always, so the baseline-backed test_new_weights is unchanged; where the ranking does change only the *order* of the exported nominal Jet collection moves, not the values. The Run2 legacy-comparison reference (comparison_arrays/output_run2.coffea) is regenerated to the new order via regenerate_run2_reference.py, which first verifies the change is coherent (same events, new nominal jets are pt-descending, and equal to the reference values re-sorted by corrected pt within 1e-5) before updating the reordered nominal Jet columns; the test keeps its order-sensitive comparison. Variation histograms are unchanged (already sorted). 

Tests: new test_jets_sort_and_jetidx_remap asserts nominal+varied jets are sorted, carry the permutation, and that jets_in_original_order restores the NanoAOD order so the jetIdx b-tag lookup matches the original jet. test_calibrators, test_new_weights, test_shape_variations and test_columns_export pass.


valsdav:

PR #535, opened 2026-07-08
Propagate JES/JER to type-1 MET via invariant pt_raw

valsdav:
The type-1 MET recompute derives each jet's raw pt and JEC factor from rawFactor. But
JetsCalibrator.apply_variation swapped only pt/mass to the varied values and left rawFactor
at its nominal value, so for a JES/JER variation the reconstructed raw pt became
pt_raw*(varied/nominal) and the L1 term of the type-1 correction was scaled by
varied/nominal, mis-propagating the variation into the MET (the nominal MET was fine).

Fix it at the source: apply_variation now recomputes rawFactor = 1 - pt_raw/pt for the
varied jets, exactly as initialize() does for the nominal. The raw pt (pt_raw) is the
uncorrected pt computed once from the nominal NanoAOD in jet_correction_corrlib and is
invariant under the JES/JER variations (verified: per-event multiset of pt_raw is identical
across nominal / JES / JER while the varied pt differs). With a consistent rawFactor,
1/(1-rawFactor) is the correct (possibly varied) L1L2L3 factor and pt*(1-rawFactor) is the
invariant raw pt, so the METCalibrator keeps its original formula with no special-casing.
This also makes rawFactor itself correct for every consumer of the varied jets, not just
the MET.

The nominal MET is unchanged (apply_variation is not called for the nominal, whose
rawFactor already comes from initialize).

### DIFF
--- a/pocket_coffea/lib/calibrators/common/common.py
+++ b/pocket_coffea/lib/calibrators/common/common.py
@@ -355,6 +355,17 @@ class JetsCalibrator(Calibrator):
             out[jet_coll_name]["pt"] = self.jets_calibrated[jet_coll_name][f"pt_{variation_type}_down"]
             out[jet_coll_name]["mass"] = self.jets_calibrated[jet_coll_name][f"mass_{variation_type}_down"]
 
+        # Keep the rawFactor consistent with the varied pt. The raw pt is invariant, so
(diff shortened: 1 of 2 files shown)


valsdav:

PR #536, opened 2026-07-08
Cache the histogram-fill hot path (broadcast, masks, data structure)     

valsdav:
HistManager.fill_histograms recomputed several per-(category,subsample) quantities for
every histogram. The biggest one — the per-object weight broadcast for 2D collection
histograms (flatten(data_structure * weight[mask])) — was explicitly NOT cached by the
weights_cache decorator, so it reran for every histogram sharing a collection.  For example when accumulating Jet.pt.

- The weights_cache decorator now caches the broadcast for every per-event (mask.ndim==1)
  fill, including the ndim==2 data_structure (collection) case. The caller makes the cache
  key collection-specific (`category::coll::<coll>`) so histograms on the same collection
  reuse the broadcast while different collections never collide. A 2D mask (cut on the
  collection) is still not cached.
- The (category, subsample) event masks and their emptiness (`cat & subs`, `ak.sum==0`)
  are precomputed once before the histogram loop instead of per histogram.
- `ak.ones_like(masked_collection)` (the broadcast data structure) is cached per
  (collection, category, subsample).
- The `weight * weight_sub` product is skipped when there are no subsamples (it was a
  no-op `* 1.` copy).

Output-identical: this is pure caching/reordering (the only arithmetic change, `x*1.0`
-> `x`, is exact). Verified BIT-IDENTICAL (np.array_equal over every histogram value and
variance with flow, plus sumw/sumw2/cutflow) between before and after on the
test_new_weights config, and the reference-backed test_new_weights /
test_shape_variations subsample+variation tests still pass.

Profiling (test_new_weights, one 2018 TTTo2L2Nu chunk, cProfile):
  fill_histograms cumtime      15.55s -> 6.79s   (2.3x)
  broadcast computations         720 -> 270 calls (947 requests, 677 now cache hits)
  get_masks calls                168 -> 20
  ones_like calls                138 -> 48
The speedup grows with the number of 2D collection histograms and weight variations.

### DIFF
--- a/pocket_coffea/lib/hist_manager.py
+++ b/pocket_coffea/lib/hist_manager.py
@@ -486,7 +486,20 @@ class HistManager:
                         # General collections
                         if ax.pos == None:
                             data = events[ax.coll][ax.field]
-                            # pos==None on a collection is the only source of ndim>1 data
+                            # pos==None on a collection is the only source of ndim>1 data.
+                            # All ndim>1 axes of a histogram must come from the SAME
+                            # collection: the broadcast/data-structure caches assume a
+                            # single per-event layout keyed by `data_coll`, and two
+                            # collections may have different jagged counts per event (which
+                            # would also break the shared flatten/fill). The data_ndim check
+                            # below does NOT catch this since both collections are ndim==2.
+                            if data_coll is not None and data_coll != ax.coll:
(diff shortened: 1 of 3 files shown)


valsdav:

Issue #537, opened 2026-07-08
Plotting performance improvements


valsdav:
Plotting large collections of histograms has a large memory footprint at the moment. This is because the full coffea output is loaded in memory before even starting producing figures.  
At the moment this is hard to improve as the coffea file format does not allow to read only one histogram at a time. 

Planning a refactoring of the histogram loading code in the plotter. 

- [ ] Save a new format "by variable" to be able to parallelize plotting and reduce memory footprint

felixzinn:
what is the timescale of this issue?
#511 is currently WIP to upgrade to the major mplhep version 1.0
should we move to the new version before or after?

valsdav:
Hi Felix, for the moment the performance improvemens I prepared are just related to the loading/storing of the histograms in the files to be able to load huge outputs without saturating RAM. So I think the 2 can proceed in parallel, they should ne be conflicting.

PR #538, opened 2026-07-08
New output splitting for more efficient plotting

valsdav:
Addresses:  #537 

Plotting large outputs consumed gigabytes of RAM: coffea's monolithic cloudpickle .coffea must be fully deserialized to read anything, and the plotter then eagerly built and retained a Shape (with deepcopies + per-category stack/systematic caches) for every (variable, year), while multiprocessing re-pickled the whole PlotManager per task.

This introduces an alternative *on-disk format* that stores each variable in its own member so histograms can be loaded one variable at a time, plus a lazy, streaming PlotManager that builds/plots/drops one Shape at a time. Peak plotting memory becomes ~O(workers) shapes instead of O(all shapes).

- output_split.py: split-format writer/reader (zip container, .coffea name, auto-detected by magic bytes) with format.json index + a small metadata member (readable without histograms) + per-variable members. 
- convert-output CLI: convert monolithic <-> split (and merge several inputs into one split file).
- merge-outputs: --split / --output-format to export the split format directly.
- plot_utils.PlotManager: consume a provider and build Shapes lazily via _build_shape(); module-level workers + a pre-fork copy-on-write global instead of pickling the manager per task; --no-cache now also bounds the systematics cache. Constructor stays backward compatible (hist_objs=/datasets_metadata=).
- make_plots: build a provider, drop the load double-buffer, prune to requested variables early. load_output: auto-detect either format.
- Existing monolithic .coffea files **still plot unchanged**.
- Tests: tests/test_output_split.py (format round-trip, metadata-only reads, providers, convert/merge) and tests/test_plotting.py (mono-vs-split plot equivalence, serial-vs-fork, --no-cache, --only-hist, lazy construction).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,6 +164,7 @@ cython_debug/
 AnalysisConfigs/
 VHccPoCo/
 output*
+!pocket_coffea/utils/output_split.py
 plots*
 .pkl.gz
 myenv/
--- a/pocket_coffea/__main__.py
+++ b/pocket_coffea/__main__.py
@@ -2,7 +2,7 @@ import click
 from rich import print
 
 import pocket_coffea
-from pocket_coffea.scripts import merge_columns, merge_outputs, split_output
+from pocket_coffea.scripts import convert_output, merge_columns, merge_outputs, split_output
 from pocket_coffea.scripts.check_jobs import check_jobs
 from pocket_coffea.scripts.dataset.build_datasets import build_datasets
 from pocket_coffea.scripts.dataset.dataset_query import dataset_discovery_cli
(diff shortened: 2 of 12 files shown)


felixzinn:
Hi,

so to make sure I understand this correctly
the output of the processor is not changed, this PR just introduces a way to split this output into a zip compressed file so that the plotting script can load one variable at a time?

coffea itself provides a compression argument to `.save` (https://coffea-hep.readthedocs.io/en/v0.7.31/api/coffea.util.save.html)
so it might be worth to distinguish that in the code, I saw that in some functions a `compression` argument is used

and what about the documentation? this introduces a new script correct? I think it would be beneficial to have an explicit documentation of the methods introduced in this PR

another thought, maybe coffea itself would be open to changes in the .coffea file format and partial loading 🤔

PR #540, opened 2026-07-09
Cache parsed correctionlib CorrectionSets by file path

valsdav:
Every sf_* helper and the jet calibrators call correctionlib.CorrectionSet.from_file (or parse_raw for JER) inside the per-chunk, per-variation call path, so the multi-MB BTV / JERC / lepton-SF / pileup JSONs are re-parsed once per (chunk x shape variation) even though the parsed evaluator is immutable and reusable.

Add lib/correction_cache.load_correction_set(path): a process-local functools.lru_cache over CorrectionSet.from_file, keyed by the resolved path. The path already encodes the year/era (params[...][year][...]["file"]), so chunks with different metadata resolve to different files and get the right one, while chunks/variations sharing a file hit the cache. It is unbounded (the distinct-file set is small) and per worker process (from_file objects are never pickled). Route all 24 from_file sites (scale_factors, leptons, jets, cut_functions, the muon calibrator, weights_run2_UL) through it.

The JER set is filtered in place, so instead of caching the raw parsed schema, jets.get_jer_correction_set (which parses, filters a *local* schema, and returns a read-only to_evaluator()) is itself memoized by (jer_json, jer_tags) -- the caller already passes jer_tags as a tuple. No mutated object is shared.

Behavior-neutral: load_correction_set returns the same evaluator from_file would, so SF values are identical. test_new_weights (uses sf_btag, pileup, ele/mu SFs) passes unchanged; new offline test_correction_cache verifies parse-once-per-path.

Profiling (config_shape_var_and_weights_bysubsample, MC 2018 with shape variations, 3 chunks, cProfile) -- the total JSON parse time drops from 30.1s to 7.1s (~23s, 4.2x):
  CorrectionSet.from_file   198 calls / 10.1s  ->  6 calls / 1.9s
  parse_raw (multi-MB JERC)   9 calls / 20.0s  ->  2 calls / 5.3s
The same files are parsed once per worker instead of once per (chunk x shape variation), so
the saving grows ~linearly with chunks x shape variations: a production job pays the parse
cost essentially once instead of ~10s per chunk.

### DIFF
--- a/pocket_coffea/lib/calibrators/common/common.py
+++ b/pocket_coffea/lib/calibrators/common/common.py
@@ -16,6 +16,7 @@ from omegaconf import OmegaConf
 from pocket_coffea.lib.muon_scale_and_resolution import pt_scale, pt_resol, pt_scale_var, pt_resol_var
 from pocket_coffea.utils.utils import get_nano_version
 import correctionlib
+from pocket_coffea.lib.correction_cache import load_correction_set
 
 class JetsCalibrator(Calibrator):
     """
@@ -823,7 +824,7 @@ class MuonsCalibrator(Calibrator):
             return
 
         self.enabled = True
-        self.cset = correctionlib.CorrectionSet.from_file(
+        self.cset = load_correction_set(
             self.mscare_params.correctionlib_config[self._year]["file"]
         )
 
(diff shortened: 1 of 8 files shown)


valsdav:

PR #541, opened 2026-07-09
Consolidate manual-job resubmission into check-jobs; drop run --recreate-jobs

valsdav:
The runner/executor `--recreate-jobs` path and `check-jobs --resubmit` duplicated a lot of logic (two `find_other_file`, two queue-bump helpers + two `queues` lists, two fileset-rewrite + condor-resubmit loops) while operating on the same jobs_dir on-disk contract. Fold all of it into `check-jobs` so there is a single consumer, and remove the runner path.

check-jobs gains a one-shot proactive recreate mode and the recreate-only knobs that used to live on `pocket-coffea run`:
- `--recreate auto|0,1,3` (+ `--once`) — recreate/resubmit a chosen set of   jobs and exit; unlike `--resubmit` it can act on running/idle jobs too.
- `--use-redirector`, `--blocklist-sites` (unioned with the auto blacklist),    `--recreate-queue`, `--skip-bad-files`.

Shared/reused instead of duplicated:
- new `utils/htcondor_queue.py` (QUEUES, bump_queue, set_queue) replaces   check_jobs.bump_jobqueue and lxplus update_queue/set_queue + both `queues`.
- `site_rewrite.find_other_file` extended (exclude_urls, blocklist-by-prefix,   fallback_redirector=None) and now used by check-jobs in place of its private copy.
- the submit-time inner_run_options helpers stay in executors_manual_jobs and are reused by check-jobs' `--skip-bad-files`.

Removed: recreate_jobs from executors_lxplus/rubin, the ABC abstractmethod, the submit() dispatch and __init__ guard-skip, and the  --blocklist-sites/--recreate-queue/--use-redirector options on the runner (--skip-bad-files stays; it also feeds the outer Coffea Runner). `run` now errors with a pointer to check-jobs if a moved flag is passed.

Also fixes two latent bugs: the babysitter's undefined `log_file` NameError on the "failed-again after resubmit" path, and the old explicit-list recreate crashing on an undefined `runningjobs`.


valsdav:
What do you think about this interface change? Asking specifically to @mmarchegiani @mondalspandan as they introduced the two code paths

mondalspandan:
Hi Davide, did you already test these changes in an actual workflow? If not, I suggest one of us does it, and then we merge, since these are substantial changes.

PR #542, opened 2026-07-09
Vectorize muon-smearing RNG and fix per-muon seed correlation

valsdav:
get_rndm previously built one numpy MT19937 generator per muon inside two Python list comprehensions, seeded by a scalar custom SeedSequence. This was both a per-event hot spot and, more importantly, it had a small issue: SeedSequence.generate(1) used only the event number,  so every muon in the same event drew the *same* uniform for the Crystal-Ball inverse-CDF smearing.

Replace the loops with a fully vectorized 64-bit hash (_hashed_uniform, built on a vectorized splitmix64 finalizer) that mixes all three words -- event, lumi and phi_int -- so each muon gets an independent, reproducible uniform in [0,1). The dead SeedSequence class and _get_rnd_func helper are removed; the rnd_gen argument is kept for backward compatibility but no longer selects a generator.

Muons in the same event are now smeared independently.  Added offline unit tests (tests/test_muon_rng.py) asserting range, reproducibility, per-muon independence (the bug fix), uniformity and splitmix64 avalanche.

### DIFF
--- a/pocket_coffea/lib/muon_scale_and_resolution.py
+++ b/pocket_coffea/lib/muon_scale_and_resolution.py
@@ -5,36 +5,33 @@ import math
 from scipy.special import erfinv, erf
 from random import random
 import awkward as ak
-from typing import List
 
 import warnings
 warnings.filterwarnings("ignore", category=RuntimeWarning)
 
-class SeedSequence:
-    def __init__(self, seeds: List[int]):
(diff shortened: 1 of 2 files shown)


valsdav:
Closing this: we are going to migrate to the pure correctionlib implementation: 

https://gitlab.cern.ch/cms-muonPOG/muonscarekit/-/blob/master/scripts/apply_corrections_coffea.py?ref_type=heads

PR #543, opened 2026-07-11
 Use official Muon POG RandomSmearing (hashprng) for muon-smearing RNG

valsdav:
Replace the hand-rolled muon-smearing RNG in muon_scale_and_resolution.py with
the current official upstream MuonScaRe.py, which draws the per-muon uniform
from the "RandomSmearing" correction (a correctionlib hashprng node keyed on
event/lumi/phi) shipped inside muon_scalesmearing.json.gz. 

MuonsCalibrator drops the now-removed rnd_gen="np" argument. Scale/resolution
ordering and variation wiring already matched the official example.

Rewrite tests/test_muon_rng.py to guard the same properties (reproducibility,
per-muon and per-event independence, uniformity) through the new correctionlib
path


valsdav:

PR #548, opened 2026-07-20
Update DESY executor and fix loop over merged jet collections for variations

andreypz:
This is to update Parsl executor at DESY, and some minor changes to the one for RWTH.

Some print out statements added for debugging in jet calibrations (could be removed if that's preferred.)


andreypz:
Updated this PR, adding one line related to variations of custom jet calibrators merging with `AK4CorrT1METJetPuppi` collection. Namely [here](https://github.com/PocketCoffea/PocketCoffea/pull/548/changes#diff-2019ed578769e72f5596d3a1e6751e00cbee4b0f94f60a6de944304adf9fd9f6L168).
I think this is actually a logical bug. The only reason why the code worked for nominal jets is because the name of the *alias* for *AK4CorrT1METJetPuppi* is `AK4PFPuppi`, which is also a part of `merge_collections_for_variations` list.
With the small change, there should be no difference for nominal jets.

Other changes to this file are just debug statements (all commented out).

felixzinn:
hi,
why not simply remove the debug statements entirely?

andreypz:
I could remove the debugs if that's preferred. Personally I prefer to keep them. I suspect it is not the last time I will need to debug this part of the code. I'm still not 100% sure it does what it's meant to do when it comes to merging variations.

pocket_coffea/executors/executors_DESY_NAF.py
```diff
@@ -68,21 +73,19 @@ def setup(self):
...
-                            worker_init    = "\n".join(self.get_worker_env()),
-                            walltime       = self.run_options["walltime"],
-                            requirements   = self.run_options.get("requirements", ""),
+                            cores_per_slot  = self.run_options.get("cores-per-worker", 1),
+                            mem_per_slot    = self.run_options.get("mem-per-worker", 4),
+                            init_blocks     = self.run_options["scaleout"],
+                            max_blocks      = self.run_options["scaleout"],
+                            worker_init     = "\n".join(self.get_worker_env()),
+                            requirements    = self.run_options.get("requirements", ""),
+                            scheduler_options = "Request_Disk = %i \n+RequestRuntime = %i" % (int(self.run_options["disk-per-worker"][:-2]) * 1024 * 1024, walltime_to_seconds(self.run_options["walltime"])),
```
felixzinn, line 83: IMO an f-string would be nicer because it is more readable
```
f"Request_Disk = {(self.run_options["disk-per-worker"][:-2] * 1024 * 1024):.0f}\n+RequestRuntime = {walltime_to_seconds(self.run_options["walltime"]):.0f}"
```

pocket_coffea/executors/executors_DESY_NAF.py
```diff
@@ -4,21 +4,22 @@
...
-#from parsl.channels import LocalChannel
 from parsl.config import Config
 from parsl.executors import HighThroughputExecutor
-from parsl.launchers import SrunLauncher, SingleNodeLauncher
-from parsl.addresses import address_by_hostname, address_by_query
+from parsl.addresses import address_by_hostname
 
+def walltime_to_seconds(walltime: str) -> int:
+    h, m, s = map(int, walltime.split(":"))
+    return h * 3600 + m * 60 + s
```
felixzinn, line 16: I dont use the DESY executor, so I am curious. Is it convention to express the walltime always in the h:m:s format?

andreypz: Well, this convention come from Dask settings in pocket-coffea, so I implemented a conversion. At DESY the Runtime is actually used (default is 6 hours). On other sites (like RWTH) there is no default (runtime is unlimited) as far as I know.

pocket_coffea/executors/executors_DESY_NAF.py
andreypz: Ok, I converted this to an f-string

andreypz:
@valsdav I think this can be merged

PR #556, opened 2026-08-06
Split jet pT regression by b-tag and add per-tagger Run3 b-tagging WPs

matteomalucchi:
## Summary
Adds a reusable helper to combine jet pT regression collections (with and without neutrinos) with the standard JEC jets, choosing per jet on a configurable b-tag cut, plus per-tagger b-tagging working points for Run3.

## Changes
- **`pocket_coffea/lib/jets.py`** — add `merge_regressed_jets` (and supporting helpers) to split the pT regression by b-tag: `+neutrino` regression for high-b-tag jets, plain regression for the rest, falling back to the standard JEC jets. Works with WP names or raw b-tag scores and a selectable tagger.
- **`pocket_coffea/parameters/btagging.yaml`** — per-tagger working points for Run3 (`2022_preEE`…`2023_postBPix`), defaulting to `btagPNetB`; older years unchanged.
- **`pocket_coffea/parameters/jet_scale_factors.yaml`** — match the Run3 default b-tag shape SF to `btagPNetB` (`particleNet_shape`).
- **`pocket_coffea/lib/cut_functions.py`** — route the b-tag WP resolution through the shared helper.
- **`tests/test_merge_regressed_jets.py`** — unit tests covering all b-tag cut configurations.
- **`docs/recipes.md`** — document the b-tag split of the regression with/without neutrinos.
- Fix a `KeyError` in `get_jersmear_SFunc` for JER SFs carrying a systematic input.

### DIFF
--- a/pocket_coffea/lib/jets.py
+++ b/pocket_coffea/lib/jets.py
@@ -305,18 +305,6 @@ def btagging(Jet, btag, wp, veto=False):
         return Jet[Jet[tagger] > threshold]
 
 
-# NanoAOD b-tag discriminant branch -> the BTV correctionlib correction that
-# stores its working-point cut values. The BTV ``btagging.json.gz`` for a campaign
-# carries one ``<tagger>_wp_values`` correction per tagger, evaluated with the WP
-# name ("L"/"M"/"T"/...) to return the discriminant cut.
-_BTAG_WP_VALUES_CORRECTION = {
-    "btagDeepFlavB": "deepJet_wp_values",
-    "btagPNetB": "particleNet_wp_values",
-    "btagRobustParTAK4": "robustParticleTransformer_wp_values",
-    "btagUParTAK4": "UParTAK4_wp_values",
-}
-
-
 def get_btag_wp_score(params, year, wp, tagger):
     """Read a b-tag working-point discriminant cut from the BTV correctionlib JSON.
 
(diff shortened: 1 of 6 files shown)


valsdav:
hi @matteomalucchi ! Since you are adding a function to check the WP of BTV, I was wondering if we can start using the correctionlib file content to read the WP of the taggers instead of saving the numbers in the pocketCoffea parameters. It would be a minimal change, what would you think?

valsdav:
@andreypz would you have a look if this is find for you ?

pocket_coffea/lib/jets.py
```diff
@@ -305,26 +305,75 @@ def btagging(Jet, btag, wp, veto=False):
         return Jet[Jet[tagger] > threshold]
 
 
+# NanoAOD b-tag discriminant branch -> the BTV correctionlib correction that
+# stores its working-point cut values. The BTV ``btagging.json.gz`` for a campaign
+# carries one ``<tagger>_wp_values`` correction per tagger, evaluated with the WP
+# name ("L"/"M"/"T"/...) to return the discriminant cut.
+_BTAG_WP_VALUES_CORRECTION = {
```
valsdav: sorry for being pedantic, but I would put these values as part of the yaml parameters

andreypz:
> @andreypz would you have a look if this is find for you ?

I can test this but only next week Wednesday the earliest.

andreypz:
I tested this PR on our VHcc workflow and it's fine to merge. It does not break our workflow, and that's the main criterion. That said, we do not use the features in this PR as we simply use regression of all the jets. But the selection feature is presumably tested by @matteomalucchi

pocket_coffea/parameters/btagging.yaml
```diff
@@ -159,3 +220,10 @@ btagging:
         T: 0.4648
         XT: 0.6298
         XXT: 0.9739
+  # Map NanoAOD b-tag discriminant branch -> BTV correctionlib correction name that
+  # stores its working-point cut values (evaluated with WP name "L"/"M"/"T"/...).
+  btag_wp_values_correction:
```
valsdav: Last comment: why this is called ".._correction"?

matteomalucchi: Because it is taking it from the correction file, but true that it doesn't really make sense. Do I rename it `btag_wp_values_map`?

valsdav: perfect, thanks

PR #570, opened 2026-09-15
Update UParTAK4 regression names and recommendations

matteomalucchi:
Enhance the documentation for ParticleNet and UParTAK4 regression algorithms, providing detailed recommendations for their usage across different years and b-tagging scenarios. Update references in the code to reflect the new naming conventions for UParTAK4.


docs/recipes.md
```diff
@@ -1150,6 +1150,21 @@ jets_calibration:
       AK4PFPuppiPNetRegression: True
       #AK4PFPuppiPNetRegressionPlusNeutrino: True
 ```
+
+Both the ParticleNet (PNet) and the UParTAK4 regression are available as `jet_types` for every year. PNet is recommended for 2022/2023, while UParTAK4 is recommended for Run 2 and for 2024/2025. For each algorithm two variants exist -- with and without neutrinos used in the training -- named by the corresponding `jet_types`/`collection` tag: the variant without neutrinos is recommended for jets below the b-tag working point (low b-tag), while the variant with neutrinos is recommended for jets above the b-tag working point (high b-tag).
```
felixzinn: hi,

I think we should link to the corresponding CMS pages (probably BTV ?) with all the information about the taggers, how they are trained and what the recommendation is. These might change and we would need to keep track of that.
In the docs for PocketCoffea I would only put information about the actual implementation and the defaults?

